PR #4, opened 2021-02-16
ESP8266 guide fixes + some typos

fnune:
- Refer to the ESP8266_RTOS_SDK directory as `IDF_PATH`: this is how their documentation refers to it, and the project depends on that env var being defined.
- Remind users to create `memfault_platform_config.h`.
- Fix some typos.


fnune:
It may also be worth mentioning that users that picked the default partition table won't have an OTA slot and their devices will go into a crash loop once they flash Memfault onto it. Here's how it went for me: https://github.com/fnune/toastboard/blob/master/NOTES.md#reboot-loop-after-integrating-the-sdk

chrisc11 (approved):

PR #13, opened 2021-02-25
Support ?embed mode during SSR too

fnune:
This wasn't happening during development because the local server
doesn't do SSR like the production one does (bad idea IMO, having a
local server that does the same as production would be great).

I noticed the header, footer and sidebar were flashing for a little bit
on the Memfault app, and I thought this might be because of SSR, and how
`EmbedModeProvider` doesn't support it.

With this change, `EmbedModeProvider` gets the search params from the
React router instead of from `document`, which should work both during
SSR as well as in the browser.

### DIFF
--- a/src/components/EmbedMode.js
+++ b/src/components/EmbedMode.js
@@ -1,17 +1,16 @@
 import React, { useMemo } from "react";
+import { useLocation } from "react-router";
 
-const DEFAULTS = { enabled: false, settings: {} };
-
-const EmbedModeContext = React.createContext(DEFAULTS);
+const EmbedModeContext = React.createContext({ enabled: false, settings: {} });
 
 /** Support for embedding the docs in an iframe from our application */
 export function EmbedModeProvider({ children }) {
+    const { search } = useLocation();
     const value = useMemo(() => {
-        if (typeof window === "undefined") return DEFAULTS;
-
-        const params = new URL(document.location).searchParams;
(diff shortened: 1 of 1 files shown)


src/components/EmbedMode.js
```diff
@@ -1,17 +1,24 @@
...
+const EmbedModeContext = React.createContext({ enabled: false, settings: {} });
 
-const EmbedModeContext = React.createContext(DEFAULTS);
+function getUrlSearchParams(search) {
+    if (typeof window === "undefined") {
+        return new NodeURLSearchParams(search);
+    } else {
+        return new window.URLSearchParams(search);
+    }
+}
```
sirlantis, line 13: `ExecutionEnvironment.canUseDOM` seems to be the idiomatic Docusaurus way.

```suggestion
function getUrlSearchParams(search) {
    if (ExecutionEnvironment.canUseDOM) {
        return new URLSearchParams(search);
    } else {
        return new NodeURLSearchParams(search);
    }
}
```

sirlantis (approved):

src/components/EmbedMode.js
fnune: Thanks for the investigation. I resorted to just `search.includes("embed")`, but I'm thinking soon we'll have more params and we'll need something anyway. Maybe import some light library that can parse queries isomorphically.

PR #15, opened 2021-03-02
Use a separate base path for embed mode instead of ?embed

fnune:
This allows us to know during SSR whether the query is from our embedded
iframe, since during SSR we don't have access to query params.


sirlantis (approved):

fnune:
Thanks! Will merge once the app PR deploys.

PR #24, opened 2021-03-10
Add prettier as dev dependency

martijnthe:
I added prettier as dev dependency.
Added the same .rc file and `lint:...` package.json scripts we've got in the monorepo.
I did not add any CI scripts whatsoever. I'm not sure whether / how that would jive with Forestry.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ yarn-error.log*
 !.yarn/versions
 .pnp.*
 
+.idea/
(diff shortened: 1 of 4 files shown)


sirlantis (approved):

sirlantis:
Don't know, but I assume https://github.com/github/super-linter could help with Prettier/Forestry.

martijnthe:
> Don't know, but I assume https://github.com/github/super-linter could help with Prettier/Forestry.

I'll give it a shot.

PR #25, opened 2021-03-10
Run prettier in GH workflow

martijnthe:


martijnthe:
Failed check run from before reformatting to show that the CI check is working:
https://github.com/memfault/memfault-docs/pull/25/checks?check_run_id=2075083507

martijnthe:
@sirlantis I ended up not using Super-Linter like you suggested because it uses markdownlint instead of prettier to format .md files.

sirlantis:
Ah I see. Prettier is only used for JS (and even that is a bit undocumented) 

Thanks,
Marcel



—
Marcel Jackwerth

--- original message ---
On 10. March 2021, 9:55 AM CET notifications@github.com wrote:



@sirlantis I ended up not using Super-Linter like you suggested because it uses markdownlint instead of prettier to format .md files.

—

You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub, or unsubscribe.
--- end of original message ---

.github/workflows/lint.yml
```diff
@@ -0,0 +1,27 @@
...
+        uses: actions/checkout@v2.3.4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v2.1.5
+
+      - name: Install Dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Lint Prettier
+        run: yarn lint:prettier
```
sirlantis, line 24: For Forestry it would probably nice if the action just fixed the formatting instead of complaining about it.

https://github.com/marketplace/actions/prettier-action

blog/2019-07-23-schedules-threadx-releases-more.md
```diff
@@ -12,9 +12,7 @@ Cohort Settings card.
 
 ![/img/blog/2019-07-23-update-schedules.png](/img/blog/2019-07-23-update-schedules.png)
 
-:::note
-Times are in **UTC**.
-:::
+:::note Times are in **UTC**. :::
```
sirlantis, line 15: This breaks. 

You need to keep prosewrap as "preserve".

Prettier uses that as a default especially because of Markdown extensions.

> By default, Prettier will wrap markdown text as-is since some services use a linebreak-sensitive renderer, e.g. GitHub comment and BitBucket.

sirlantis (changes requested):
Need to get rid of prosewrap.

sirlantis (approved):
🙏🏼

martijnthe:
> Need to get rid of prosewrap.

It is a bit of a bummer that lines are now no longer auto-wrapped.

I looks like prettier has way to ignore parts in markdown:
https://prettier.io/docs/en/ignore.html#range-ignore

```
<!-- prettier-ignore-start -->
:::note
Lorem ipsum
:::
<!-- prettier-ignore-end -->
```

Would that be too much of a hassle?

PR #38, opened 2021-03-21
Adjustments to embedded getting started guide

chrisc11:
* Update instructions to recommend copying ports/template folder 
* Add a few more test commands that are good to run as a validation check
* Simplify coredump dependency setup steps and & provide links to flash coredump ports
* Rejig reboot tracking template to match what we have done when a port exists (i.e call `memfault_platform_reboot_tracking_boot()`)


martijnthe (approved):

mrfirmware (approved):
Since C11 we can use alignas. Should we use it on s_reboot_tracking or is it too redundant with the linker allocation. Although I did not see an ALIGN in the linker section either.

mrfirmware (approved):

chrisc11:
> Since C11 we can use alignas. Should we use it on s_reboot_tracking or is it too redundant with the linker allocation. Although I did not see an ALIGN in the linker section either.

We could (there's a MEMFAULT_ALIGNED macro as well) but there's no hard requirement it's aligned. The allocation gets cast to a packed struct so the compiler will emit instructions that are safe along unaligned boundaries.

PR #39, opened 2021-03-28
Improve documentation around exporting memfault chunks via CLI

chrisc11:
* Move bulk of gdb and cli based upload guides into MDX files so the steps can be dropped directly into the Cortex-M getting started guide
* Hide "memfault_data_export_dump_chunks()" override option behind toggle since 99% of folks should not need it

### DIFF
--- a/docs/embedded/arm-cortex-m-guide.mdx
+++ b/docs/embedded/arm-cortex-m-guide.mdx
@@ -107,7 +107,7 @@ instrumentation, and test the integration!
 
 <AddCliTestCommands />
 
-### Post data Locally
+### Post Data to Cloud via Local Debug Setup
 
 <PostChunksLocally />
 
@@ -124,6 +124,6 @@ instrumentation, and test the integration!
 With data collection verified locally, the final step is to push data
 automatically to the Memfault cloud for analysis.
 
-### Push Data to Cloud over Device Transport
+### Post Data to Cloud via Device Transport
 
 <PublishDataToCloud />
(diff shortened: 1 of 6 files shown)


docs/embedded/arm-cortex-m-guide.mdx
```diff
@@ -107,7 +107,7 @@ instrumentation, and test the integration!
 
 <AddCliTestCommands />
 
-### Post data Locally
+### Post Data To Memfault cloud
```
martijnthe, line 110: To me, this heading doesn't convey that this section is about posting w/o the device transport.
The first sentence explains it, but it would be good if the heading also captures that.

Maybe...

Post Data To Memfault cloud without (device) transport

Post Data To Memfault cloud for testing

Post Data To Memfault cloud out-of-band

Post Data To Memfault cloud via (serial) console or GDB

docs/embedded/steps/post-chunks-locally.mdx
```diff
@@ -8,24 +11,31 @@ local QA testing or in CI/CD test automation setups.
...
+
+#### Post Chunks To Memfault
+
+All data collected by Memfault can be exported as opaque packets called
+"chunks". To trigger an export, call the `test_export` command added to your
+device CLI or add a periodic call to `memfault_data_export_dump_chunks` in your
+code.
 
-To test it out, call the `test_export` command added to your CLI above:
+<details>
```
martijnthe, line 28: OT: I wish the details fold-out would have some visual clue as to where the end of the fold out is.

martijnthe (approved):

docs/embedded/steps/post-chunks-locally.mdx
chrisc11: Agree, hoping we can do something smarter with mdx/js for these. Also wish I could send a direct link that auto expands these in some way

docs/embedded/arm-cortex-m-guide.mdx
chrisc11: 👍 Going with "Post Data to Cloud via Local Debug Setup" in contrast to the later "Post Data to Cloud via Device Transport" step for actual end-to-end

PR #43, opened 2021-04-03
Only perform prettier check on PRs from forked repo

chrisc11:
The workflow we had configured for running prettier and auto-pushing a fix commit does not work when the pull request is generated against a forked repo for two reasons:
* "github.head_ref" is not a valid reference for the forked repo
* no clean way to push a auto-fix commit to the branch on the forked repo

I updated the lint work flow to be check-only for forked PRs so there isn't always a CI failure in this scenario and added some notes to the README about how to fix lint errors locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,14 @@ Start a local webserver that will pick up most changes without restarting:
 $ yarn start
 ```
 
+## Linter
+
+Code is linted with prettier. To fix up files to match format, run:
+
+```
+$ ./node_modules/prettier/bin-prettier.js --write /path/to/file.mdx
+```
+
 ## Content
 
 Add a new page to the docs:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -24,6 +24,14 @@ Start a local webserver that will pick up most changes without restarting:
 $ yarn start
 ```
 
+## Linter
+
+Code is linted with prettier. To fix up files to match format, run:
+
+```
+$ ./node_modules/prettier/bin-prettier.js --write /path/to/file.mdx
```
sirlantis, line 32:
```suggestion
$ ./node_modules/.bin/prettier --write path/to/file.mdx
```

sirlantis (approved):

martijnthe:
I noticed that for commits on master, the `prettier-fork` job gets run and the `prettier` job gets skipped. I don't quite understand why though. For this job it doesn't matter I think, but it would be nice to understand why it's happening.

Screenshot from 2021-04-21 09-45-28

PR #46, opened 2021-04-08
Rename blog to changelog, add redirect

tyhoff:
We wanted the changelog to be called the changelog early on, but couldn't do it do to Docusaurus not having the ability to. It can as of a while ago, so let's do it!

Also made sure to add a redirect in Netlify to forward any old requests to the new URL.


chrisc11 (approved):
💯

PR #47, opened 2021-04-12
Start with UI based chunk posting in getting started guide

chrisc11:
Updating getting started guide to suggest uploading first chunks via UI directly. This way a user will see if there are any validation errors immediately directly in the UI. 

The UI also includes instructions about how to automate the process after the first upload is complete with the added benefit that the instructions in the UI can generate copy/paste instructions with the correct arguments already populated!


docs/embedded/steps/post-data-with-memfault-ui.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+Select "Next" to review the chunks that were identified and (optionally) update
+the "Device Serial" being used to report data:
+
+![](/binary-assets/embedded/post-chunks-ui.png)
+
+Select "Post" to submit the chunks to Memfault. The chunks will appear in the
+view directly below. Check and make sure there are no "Errors" to address.
+
+![](/binary-assets/embedded/chunks-debug-view.png)
```
martijnthe, line 35: nit: the device serial here is UI_TESTSERIAL while in the other it's MFLTQC10001

martijnthe (approved):

docs/embedded/steps/post-data-with-memfault-ui.mdx
chrisc11: good catch, fixed

PR #52, opened 2021-04-30
MFLT-3511 Initial docs for Android Structured Logs

martijnthe:


docs/android/structured-logs.mdx
```diff
@@ -0,0 +1,189 @@
...
+```
+ static_libs: [
+    "memfault-structured-log-lib"
+ ]
+```
+
+#### Add structured-log-lib to Gradle-built, Java or Kotlin apps
+
+For apps that are build outside of the AOSP build system using Gradle, you will
+need.. TODO
```
martijnthe, line 133: @sjp4 @rjcase I'm not sure what the best solution here is.

I experimented with adding `structured-log-lib` to another Android Studio / gradle project by going something along these lines:

```
// settings.gradle
...

include 'bort'
project(':bort').projectDir = file('/home/mthe/memfault/memfault/sdk/bort/MemfaultPackages')

```

and then

```
// build.gradle

dependencies {
    implementation project(':bort')
    ...
}
```

But this did not work too well, because Bort's gradle scripts construct various paths using `rootProject`. I also tried importing just the `structured-log-lib` sub-project in a similar way, but that also doesn't work because its build.gradle file depends on Bort's main one to [fill in the versions](https://github.com/memfault/memfault/blob/67822960f5022a8abe0b696fca8b323407ed87a9/sdk/bort/MemfaultPackages/structured-log-lib/build.gradle#L14-L19).

Some solution ideas:
1. Publish `structured-log-lib` to Sonatype, like we do with android-cloud -- I don't expect this lib to change that much TBH.
2. Untie the `structured-log-lib` sub-project such that it does not depend on Bort's root project (basically ditch the dynamic version build props)
3. Give instructions on how to add the .java + .aidl source files directly to customer's project

I think option 1 is the easiest from a customer DX point of view I think. It's slightly more maintenance on our part, but I think it's worth it.

Thoughts?

sjp4: This might work better using gradle's `includeBuild` (https://blog.gradle.org/introducing-composite-builds), rather than the `include`/`projectDir` syntax.

Publishing an aar would also work - but that this be a quicker fix.

docs/android/structured-logs.mdx
```diff
@@ -0,0 +1,215 @@
...
+
+When diagnosing Android device issues, logcat logs are one of the most common
+sources of information to use. Unfortunately, the signal-to-noise ratio can be
+poor. On top of that, logcat prunes logs from chatty sources, causing important
+logs to get lost.
+
+The Structured Logging feature is a complementary mechanism to log events of
+interest for debugging purposes. Your application code can leverage this to log
+crucial information and get it surfaced in the Memfault UI. The Bort SDK itself
+also provides built-in sources that log events of interest from common Android
```
rjcase, line 15: Maybe remove the "built-in" sources for now, until development is complete? (If so, then also remove the section below?)

docs/android/structured-logs.mdx
```diff
@@ -0,0 +1,215 @@
...
+
+### Add the `structured-log-lib` library
+
+The `structured-log-lib` library is a small Java library that wraps the Binder
+service that is published by MemfaultStructuredLogd. The next two sections
+describe how to add this library to an AOSP-built Java app and how to add it to
+a Gradle-built Java or Kotlin app.
+
+Note it is possible to log from native C/C++ programs, but we do not have a
+C/C++ helper library at the moment.
```
rjcase, line 104: Should we add a snippet here of how to use `MemfaultStructuredLogd` from C/C++?

rjcase (approved):
Thanks @martijnthe !

docs/android/structured-logs.mdx
```diff
@@ -0,0 +1,215 @@
...
+:::tip
+
+Each Structured Log has a "type" (string) and contains structured data in the
+form of a JSON payload. The schema of the JSON payload is up to you. That said,
+we recommend using the same schema for a given type. Future aggregation features
+may require this.
+
+:::
+
+### Structured Logs vs Logcat
```
sjp4, line 51: Should we add an entry describing the difference in how logs are viewed? (i.e. structured: events on timeline page / logcat: drill-down into logs to view

docs/android/structured-logs.mdx
```diff
@@ -0,0 +1,215 @@
...
+well and can now be used from your app's Java or Kotlin source files.
+
+### Logging API
+
+```java
+import com.memfault.bort.structuredlog.StructuredLog;
+
+// Create a Structured Log
+// with type "door.open" and JSON payload: { "foo": 123, "bar": true }
+StructuredLog.log("door.open", "{\"foo\": 123, \"bar\": true}");
```
sjp4, line 168: OT (on the API, not the docs): Have we considered providing convenience APIs for common scenarios e.g. `..String tag, Map message)` or allowing an object to be passed in, which we GSON internally `..String tag, Object message)`? The escaping here to construct an object seems a bit messy.

Edit: ah I see below you recommend using a serialization library.

docs/android/structured-logs.mdx
```diff
@@ -0,0 +1,215 @@
...
+#### Add `structured-log-lib` to Gradle-built, Java or Kotlin apps
+
+For apps that are build outside of the AOSP build system using Gradle, the
+easiest way to include `structured-log-lib` in your project is using Gradle's
+`includeBuild` API.
+
+In your project's `settings.gradle` file, add:
+
+```
+includeBuild('/path/to/bort/MemfaultPackages') {
```
sjp4, line 142: I think we will probably want to also provide this as a published `aar` on nexus at some point (many companies will be developing aosp/apps completely separately, and probably won't want to cross-reference build scripts between repos).

docs/android/structured-logs.mdx
```diff
@@ -0,0 +1,215 @@
...
+
+In case the data was invalid JSON or exceeded the 4KB limit, "Invalid Data" and
+respectively "Oversized Data" swimlanes will appear in the Device Timeline,
+containing information to help debug these issues.
+
+### Testing & Debugging
+
+By default, the MemfaultStructuredLogd daemon only flushes out received logs
+periodically or when the configured maximum log count has been reached. During
+development, this is not very practical. Therefore, on debug builds of AOSP, you
```
sjp4, line 188: Maybe make it explicit that `dump()` is no-op in release builds?

docs/android/structured-logs.mdx
```diff
@@ -0,0 +1,215 @@
...
+When you build your project, the `structured-log-lib` library will be built as
+well and can now be used from your app's Java or Kotlin source files.
+
+### Logging API
+
+```java
+import com.memfault.bort.structuredlog.StructuredLog;
+
+// Create a Structured Log
+// with type "door.open" and JSON payload: { "foo": 123, "bar": true }
```
sjp4, line 167: OT: Have we thought about whether/how in the future, we might make the message more structured, to enable e.g. showing a different-coloured event bubble on the timeline depending on one of the parameters in the message? (e.g. `log("door", "{\"open\": "true", \"bar\": true}" and we would show different colors for open/closed on the same lane, or maybe even allowing for "graph" lines like we have for `batterystats` swimlanes)

docs/android/structured-logs.mdx
```diff
@@ -0,0 +1,215 @@
...
+:::
+
+### Structured Logs vs Logcat
+
+|                              | Structured Logs                                                    | Logcat                                                            |
+| ---------------------------- | :----------------------------------------------------------------- | :---------------------------------------------------------------- |
+| Log Payload                  | JSON                                                               | Text                                                              |
+| Log Metadata                 | Timestamp, type                                                    | Timestamp, log level, tag, user, PID, TID                         |
+| Log Producers                | Only subsystems you care about                                     | Everything                                                        |
+| Intended Log Frequency       | Low-Medium (10s per hour)                                          | High (1000s per hour)                                             |
```
sjp4, line 58: Is the `10s per hour` intended to be a guide for individual log sources, or an overall guide for the entire structured logging system? (I'm imagining that e.g. wifi logging by itself could easily be 100+ events an hour if there is anything happening)

sjp4 (approved):

docs/android/structured-logs.mdx
martijnthe: If we want first-class C/C++ support, I think it's better to create a little library like the java one but then for C/C++.

Added https://www.notion.so/memfault/Add-C-library-for-MemfaultStructuredLogd-a09e4563e36a4ac081bbd72bfe9bcaa8 to Diogo's backlog.

docs/android/structured-logs.mdx
martijnthe: Yeah, we wanted to avoid dragging in a dependency on a JSON serialization lib.

docs/android/structured-logs.mdx
martijnthe: Good idea -- https://memfault.myjetbrains.com/youtrack/issue/MFLT-3525

docs/android/structured-logs.mdx
martijnthe: I noticed this doesn't work atm on master. Fixed it here: https://github.com/memfault/memfault/pull/3954

docs/android/structured-logs.mdx
martijnthe: Per log source. My concern is mostly around the Timeline UX breaking down. Not so much the underlying infra.

docs/android/structured-logs.mdx
martijnthe: Not really no, this is still TBD.

PR #54, opened 2021-05-12
Fix the legibility of the "caution" admonition

martijnthe:
- Changed the code element's background color to be 25% translucent black
- Changed the yellow background color to be orange for more contrast

Before:
Screen Shot 2021-05-11 at 9 42 15 AM

After:
Screenshot from 2021-05-12 11-13-33


fnune (approved):
Sweet!

PR #58, opened 2021-05-26
MFLT-3404: Document adding memory regions to coredump

mrfirmware:
### Summary

Write a section in coredumps.mdx to explain how users
can add specific sections to the coredump saved data.

 ### Test Plan

Review doc update for corectness by following the tutorial
and verifying expected results.

### DIFF
--- a/docs/embedded/coredumps.mdx
+++ b/docs/embedded/coredumps.mdx
@@ -42,3 +42,405 @@ implementing the required
 At the very end of saving a coredump, the `memfault_platform_reboot()` function
 implemented as part of your initial port will be called. From here any final
 system cleanup can be performed before restarting the system
+
+### Choosing what to store in a coredump
+
+For the best debugging experience it is best to store all of RAM to
+backing storage. However, this is not always possible or desirable.
+For one, there simply may not be enough storage to preserve all of
+RAM. In other cases there may be sensitive data on the device that
+should not be sent off device. In this section you will learn how to
+choose what to save and how to make this easier for unrelated sections
+of RAM in the case where you cannot or should not preserve all of RAM.
+
+:::note
+
+This tutorial assumes a `gcc` toolchain but the concepts are similar
+with other toolchains.
+
+:::
+
+Memfault coredumps are designed to be flexible allowing the user to
+choose only what they want or what they can fit into the coredump data
+storage. Memfault uses *coredump regions* to describe what regions of
+memory are to be gathered up and preserved when a fault exception or
+user trace occurs.
+
+For supported RTOSs like FreeRTOS or Zephyr Memfault automatically
+collects task information if you use the Memfault port. But what if
+you have a special block of data or set of structures in your
+application that you would like to capture as part of a coredump? What
+if you want to store all of RAM except a particular set of structures?
+
+To achieve this control we will use the linker to locate named
+sections within the memory map of your application. Using those
+section names and special compiler attributes we can assign any object
+to a given section.
+
+Once the objects are located in named sections we will add those
+sections as regions to the coredump saving logic. This last part will
+require modifying an existing version of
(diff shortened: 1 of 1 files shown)


docs/embedded/coredumps.mdx
```diff
@@ -42,3 +42,389 @@ implementing the required
...
+```c
+[...]
+MEMFAULT_PUT_IN_SECTION(".dma_buffers")
+static uin8_t s_uart_dma_buffer[NUM_UARTS][256];
+```
+
+To share a structure between your bootloader and production application the boot
+loader would need to ensure that its linker script file had a matching entry for
+the `SHAREDRAM` memory region and the `.shared` section. After that, the two
+separate applications could exchange information in well-known data structures
```
martijnthe, line 402:
```suggestion
separate applications could exchange information in common data structures
```

martijnthe (approved):

docs/embedded/coredumps.mdx
```diff
@@ -42,3 +42,389 @@ implementing the required
...
+For the best debugging experience it is best to store all of RAM to backing
+storage. However, this is not always possible or desirable. For one, there
+simply may not be enough storage to preserve all of RAM. In other cases there
+may be sensitive data on the device that should not be sent off device. In this
+section you will learn how to chose what to save and how to make this easier for
+unrelated sections of RAM in the case where you cannot or should not preserve
+all of RAM.
+
+:::note This tutorial assumes a `gcc` toolchain but the concepts are similar
+with other toolchains. :::
```
chrisc11, line 57: I think this has to be on it's own line to render correctly
image

```suggestion
:::note
This tutorial assumes a `gcc` toolchain but the concepts are similar
with other toolchains.
:::
```

docs/embedded/coredumps.mdx
```diff
@@ -42,3 +42,389 @@ implementing the required
...
+char g_name[128];
+
+int bar(void) {
+  MEMFAULT_PUT_IN_SECTION(".memfault_data")
+  static char s_num_resources = 12;
+  [...]
+}
+```
+
+If you do nothing more you will find that these variables have indeterminate
```
chrisc11, line 283: One idea: Can't we put the section in the same block as important_allocations inside bss/data and then they would still get initialized for us automatically?

```
        memfault_bss_start = .;
        *important_allocations.o(.bss COMMON .bss*)
        *(.memfault_data.*)
        memfault_bss_end = .;
```

chrisc11 (approved):

docs/embedded/coredumps.mdx
mrfirmware: I'm not sure that we can. For one, we need a way to tell the linker where to put the value initializers for the .memfault_data section and we don't do that for NOLOAD sections like .bss. Secondly, the C run-time will only `memset()` the region defined by the .bss section so we don't get an automatic init for our .data vars. Maybe I'm not understanding what you want to do here?

chrisc11: Oops, typo in my example, should have been `memfault_bss` instead of `memfault_data`:

```
        memfault_bss_start = .;
        *important_allocations.o(.bss COMMON .bss*)
=>        *(.memfault_bss.*)
        memfault_bss_end = .;
```

If the memfault_bss & memfault_data blocks were placed in the pre-existing sections, I think everything just works:

```
    .bss (NOLOAD) :
    {
        __bss_start__ = .;
        *(.bss*)
        *(COMMON)
        memfault_bss_start = .;
        *important_allocations.o(.bss COMMON .bss*)
        *(.memfault_bss.*)
        memfault_bss_end = .;
        __bss_end__ = .;
    } >RAM
```

mrfirmware: Oh, that makes sense. Yes, I can move the section into the default sections.

PR #60, opened 2021-05-31
MFLT-3548: Substitute YOUR_PROJECT_KEY in docs with actual key from URL params

fnune:
## Summary

 I ran through some iterations of this: searching with XPath, feeding
 XPath results to a `MutationObserver`, substituting `textContent` and
 `innerHTML`... everything was bad, but I decided to read the Docusaurus
 docs a bit more and found this concept of "swizzling": basically
 ejecting Docusaurus theme components and substituting them with your
 own.

 [Docs on swizzling here](https://docusaurus.io/docs/using-themes#wrapping-theme-components).

 I basically "swizzled" the `CodeBlock` component and am running
 `.replace` on children that are strings.

 The different targets `<YOUR_PROJECT_KEY>`, `${YOUR_PROJECT_KEY}` are
 necessary because when the `projectKey` variable is _not_ present, the
 preexisting strings still need to make sense in their context. For
 example,  `${YOUR_PROJECT_KEY}` would be substituted by
 `THE_PROJECT_KEY` for a bash command to work, but if we don't have the
 actual value, then we still want to show the decorations, which may
 help with interpretation in shells.

 ## Test plan

 Manual:
 - Added `?projectKey=blabla` to several pages and checked that the
   substitution works
 - Checked that navigating to other pages still runs the substitution
   even if the param is no longer present (our context saves the values)

### DIFF
--- a/docs/ci/authentication.mdx
+++ b/docs/ci/authentication.mdx
@@ -113,14 +113,14 @@ Every project member see the project's Project Key.
   <TabItem value="api">
 
 ```bash
-curl --header "Memfault-Project-Key: $PROJECT_KEY" ...
+curl --header "Memfault-Project-Key: $YOUR_PROJECT_KEY" ...
 ```
 
   </TabItem>
   <TabItem value="cli">
 
 ```bash
-memfault --project-key "$PROJECT_KEY" ...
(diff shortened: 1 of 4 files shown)


chrisc11 (approved):
Neat!

PR #61, opened 2021-05-31
MFLT-3556: Mimic main website dir structure in /embed

fnune:
I'm trying to reuse the links in the web application, which we currently
use as `src` to an `iframe`, to add a link so that users can read the
docs on `docs.memfault.com` as well.

It soon became apparent that having a different URL structure for
embed mode and normal mode was a bad idea.

The main website look like this:

- `/docs/embedded/*.mdx`
- `/docs/android/*.mdx`

However embed mode looked like this:

- `/embed/*.mdx`

So now they are both namespaced the same way, with `android` and
`embedded`.

I've also added redirects to our Netlify config so that this can be
deployed without orchestration.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -66,8 +66,7 @@ module.exports = {
                 },
 
                 {
-                    to:
-                        "https://memfault.com/register?utm_campaign=Self%20Serve%20Launch&utm_source=Docs",
+                    to: "https://memfault.com/register?utm_campaign=Self%20Serve%20Launch&utm_source=Docs",
                     label: "API",
                     position: "left",
                     target: "_blank",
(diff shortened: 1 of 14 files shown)


docusaurus.config.js
```diff
@@ -66,7 +66,8 @@ module.exports = {
                 },
 
                 {
-                    to: "https://memfault.com/register?utm_campaign=Self%20Serve%20Launch&utm_source=Docs",
+                    to:
+                        "https://memfault.com/register?utm_campaign=Self%20Serve%20Launch&utm_source=Docs",
```
fnune: My editor decided it was a good time to do this now (?)

fnune: Oh, it was a GH action

chrisc11 (approved):

PR #65, opened 2021-06-03
Update User API key docs for releases

tyhoff:
@topher200 mentioned he could have used this information when 
setting up the CLI. I've updated the location (General/Settings to Profile)
and added the note about using it in `--password`.


topher200 (commented):
:shipit:

PR #66, opened 2021-06-08
MFLT-3637: Improve nRF SDK docs: include tabs depending on SDK version

fnune:
Our nRF Connect SDK guide needed an update for the upcoming SDK version
1.6.0 by Nordic, which [ships with built-in support for Memfault](https://github.com/nrfconnect/sdk-nrf/pull/4670).

It has not been released yet, so maybe it's necessary to hold back on
merging this until then.

Additionally, I've made some other changes:

- Generally reduced drama by changing `:::caution` to `:::important`
  (cautions are orange, "importants" are blue).
- Added a `ProjectAwareCreateAProject` component that will either
  disappear if in `/embed` mode, show the project key if it's available,
  or show what we showed before otherwise ("create a project and get a
  key").


chrisc11 (approved):
lgtm!

PR #68, opened 2021-06-14
Remove accidental pages created from MDX partials

fnune:
The warning at the bottom of [this Docusaurus docs
section](https://docusaurus.io/docs/next/creating-pages#routing) says
that any and all components, Markdown files and MDX files under `docs`
will get their own page unless they're excluded with a config setting
that I couldn't find documentation for, or if their path starts with an
underscore.

Pages created accidentally look like this:

- https://docs.memfault.com/docs/embedded/steps/arch-dependencies/
- https://docs.memfault.com/docs/embedded/steps/create-project-key/

I noticed this because I'm working on the DocSearch feature and
Algolia's crawler was showing these pages as search results.


martijnthe (commented):
Good to know!

PR #69, opened 2021-06-15
Reprise: remove accidental pages created from MDX partials

fnune:
Initially attempted here:
https://github.com/memfault/memfault-docs/pull/68

I realized the pages were being created anyway.

A "fix" is to move all partials into `src/pages/_partials/*` and prepend
**all path sections** with an underscore.

To learn the "why", please check out my comments here:
https://github.com/facebook/docusaurus/pull/3196#issuecomment-861560567

## Test plan

- Generate a sitemap, check that the partials aren't there.
- Try opening the pages where they would be, check that they're aren't there.


fnune:
A patch to support `_` in more places is being tracked here https://github.com/facebook/docusaurus/issues/4984

fnune:
This is the only workaround that exists (as [confirmed by Docusaurus maintainers](https://github.com/facebook/docusaurus/pull/3196#issuecomment-862229511)) so I'm going to merge it to get it fixed ASAP.

This will bring improvements in terms of SEO (no duplicate content) and will enable us to activate the DocSearch feature by Algolia, which currently behaves weirdly because it finds all those partials and links to them.

PR #71, opened 2021-06-16
Use Inter to match the font used in the application

fnune:
image

Before on Linux (defaulting to `Ubuntu`):

image


HBehrens:
I can see that we want to establish brand recognition with our docs. How does this compare to our website?

fnune:
CircularStd:

image

IDK, it's a fitting font for landing page copy, but not for content IMO. I'm no designer, though.

chrisc11 (approved):

PR #80, opened 2021-06-18
MFLT-3742 Document how to send data to Memfault over UDP

fnune:


fnune:
@HBehrens Not sure if we need something more elaborate?

martijnthe (approved):

PR #81, opened 2021-06-29
Remove minimum version from docs

HBehrens:
Since we deprecated the use of minimum version and 
don't mention it anywhere else in our documentation, 
we remove it from the glossary, too.


tyhoff (approved):
Oh, did not realize this was still there! Yes please.

PR #86, opened 2021-07-09
Rename structured logs to custom events in docs

sjp4:
Keeps the old page, which links to the new one (just until we have done the frontend release that links to the new page).

Will not merge until Bort 3.6.0 (with the renamed APIs) is released.

Still has a figma diagram referencing Structured Logs. I'll fix that up when I get access to it.


chrisc11:
Not sure if it's what you want here but you can redirect old pages to new ones from `static/_redirects` so old links floating around will continue to work fwiw

sjp4:
> Not sure if it's what you want here but you can redirect old pages to new ones from `static/_redirects` so old links floating around will continue to work fwiw

Updated to redirect from old->new pages!

rjcase (approved):
Thanks!

PR #98, opened 2021-08-17
Add Device Attributes

sirlantis:


docs/platform/device-attributes.mdx
```diff
@@ -0,0 +1,77 @@
+---
+title: Device Attributes
+---
+
+You can store custom attributes for each of your devices devices.
```
rjcase, line 5:
```suggestion
You can store custom attributes for each of your devices.
```

docs/platform/device-attributes.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+:::info Attributes and Metrics
+
+Any metric that a device reports through the Metric SDK automatically becomes a
+device attribute (the value being the latest reading). Therefore metric and
+attribute definitions are one and the same.
```
rjcase, line 11: One and the same means exactly the same -- but the terms are not synonymous. A device's attributes are the set of most recently read metrics?

docs/platform/device-attributes.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+:::
+
+## Defining an Attribute
+
+An attribute is defined the first time you write it. You can view all defined
+properties going to **Project Settings** and then **Metrics**. An attribute
```
rjcase, line 18: It feels surprising to me that you define and edit attributes in a setting called "metrics" -- maybe we should rename that to attributes, or otherwise change these docs to say this is how you define metrics (not attributes)?

It's small but just calling out the inconsistency

docs/platform/device-attributes.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+1. using the [Metrics API](/docs/mcu/metrics-api),
+2. using the
+   [REST API](https://api-docs.memfault.com/#f4e2584a-36f2-4184-82ec-3b9d0108588c),
+   or
+3. using the web application.
+
+:::caution
+
+Because devices can update attributes, anyone who holds your non-secret
```
rjcase, line 50:
```suggestion
Because devices can update attributes, anyone who holds your 
```

You are of course technically correct, but I'm wondering if that's still something we want to call out here using that language

docs/platform/device-attributes.mdx
```diff
@@ -0,0 +1,77 @@
...
+**Filter** panel and select the **Custom Metric** you want to search by.
+
+![Example of filtering by an attribute](/img/blog/2021-08-17-device-properties-filter.png)
+
+You can search by multiple attributes using the **+** icon.
+
+:::note Searching by Attribute vs Metric Readings
+
+If you select a **During** time frame, devices will be searched by their metric
+readings during that time frame, not by their device attributes.
```
rjcase, line 75: Maybe just re-iterate the difference?
```suggestion
readings during that time frame, not by their device attributes (the most recently reported metric).
```

rjcase (approved):
Thanks!! Our usage of the terms metrics and attributes in the docs vs the app feels somewhat blurry. Not a showstopper but maybe something we want to have a quick and careful think about

docs/platform/device-attributes.mdx
HBehrens: I'd weaken this as we are planning to extend to settings per metric if they are an attribute in the first place.

> You can configure device attributes through the metric settings.

docs/platform/device-attributes.mdx
HBehrens: Yes, I keep calling this out towards @chrisc11 – we need an umbrella term that describes the structure of "aggregated readings", "readings, not stored in TSDB (future)", "last-seen values per device aka attributes"

docs/platform/device-attributes.mdx
```diff
@@ -0,0 +1,77 @@
...
+properties going to **Project Settings** and then **Metrics**. An attribute
+definition can have a few properties:
+
+- A `string_key` which devices and APIs can use to identify the attribute.
+- Whether the data type is an `INT` (default), `FLOAT`, `BOOL`, or `STRING`.
+- Whether or not the value is `updatable` (or write-once).
+- Whether it is a `favorite` attribute, which will cause it to appear more
+  prominently in a few places in the web app.
+- The value `range` which informs charts and other components.
+- A set of `ignored_values` (like `-1`) which should be ignored.
```
HBehrens, line 27: Aren't all of these values equally relevant to metrics in general? Do we have a place to put this instead so we can link from here?

docs/platform/device-attributes.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+![Example of some attribute definitions](/img/blog/2021-08-17-device-properties-settings.png)
+
+## Viewing an Attribute
+
+Find the device you want to view the attributes for (for example by browsing
+your devices on the **Devices** page) and select the **Properties** tab. You'll
+see all the properties known for this project and the respective value for the
```
HBehrens, line 35: Is this outdated? I thought we'd called this feature (incl. the tab) "Attributes"?

docs/platform/device-attributes.mdx
```diff
@@ -0,0 +1,77 @@
...
+device you're currently looking at.
+
+## Writing an Attribute
```
HBehrens, line 38: "Setting"?

docs/platform/device-attributes.mdx
```diff
@@ -0,0 +1,77 @@
...
+![Example of editing an attribute via the web application](/img/blog/2021-08-17-device-properties-edit.png)
+
+## Searching by Attribute
+
+You can search devices by attributes. Visiting the **Devices** page, open the
+**Filter** panel and select the **Custom Metric** you want to search by.
+
+![Example of filtering by an attribute](/img/blog/2021-08-17-device-properties-filter.png)
+
+You can search by multiple attributes using the **+** icon.
```
HBehrens, line 70: consistency: above we said _pencil icon_ (italic) here, it's **+** (bold)

I wonder if we should find a way to re-use the actual icons from our app in the docs…

docs/platform/device-attributes.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+:::note Searching by Attribute vs Metric Readings
+
+If you select a **During** time frame, devices will be searched by their metric
+readings during that time frame, not by their device attributes (the most
+recently reported metric).
```
HBehrens, line 76: "latest reported value"?

HBehrens (approved):
Generally looks good to me, a few nits.

docs/platform/device-attributes.mdx
```diff
@@ -0,0 +1,78 @@
...
+title: Device Attributes
+---
+
+You can store custom attributes for each of your devices devices.
+
+:::info Attributes and Metrics
+
+Any metric that a device reports through the Metric SDK automatically becomes a
+device attribute (the value being the latest reading). Therefore metric and
+attribute definitions are one and the same.
```
franc0is, line 11:
```suggestion
Any metric that a device reports through the Metric SDK automatically populates a device attribute with its latest reading. For example, sending `battery_pct` as a metric will populate a `battery_pct` attribute on the device with the last known battery level read from the device. As such, you can configure both metrics and attributes via the Metrics section of the Settings page.
```

docs/platform/device-attributes.mdx
sirlantis: @HBehrens are thinking about moving towards "Attributes" as the general term, and an attribute can be flagged to also be a "device metric" (which by default they all are). Will finetune this in a later iteration.

docs/platform/device-attributes.mdx
```diff
@@ -0,0 +1,78 @@
...
+  prominently in a few places in the web app.
+- The value `range` which informs charts and other components.
+- A set of `ignored_values` (like `-1`) which should be ignored.
+
+![Example of some attribute definitions](/img/blog/2021-08-17-device-properties-settings.png)
+
+## Viewing an Attribute
+
+Find the device you want to view the attributes for (for example by browsing
+your devices on the **Devices** page) and select the **Properties** tab. You'll
```
sirlantis, line 34:
```suggestion
your devices on the **Devices** page) and select the **Attributes** tab. You'll
```

docs/platform/device-attributes.mdx
sirlantis:
```suggestion
## Setting an Attribute
```

docs/platform/device-attributes.mdx
sirlantis:
```suggestion
latest reported metric).
```

docs/platform/device-attributes.mdx
sirlantis: As we consider making attributes the umbrella term, we would instead link here.

docs/platform/device-attributes.mdx
sirlantis: It's pretty important customers understand this and that they can not trust attributes being correct/protected from bad actors.

docs/platform/device-attributes.mdx
sirlantis:
```suggestion
You can search by multiple attributes using the _plus icon_.
```

sidebars.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
             "platform/fleet-wide-metrics",
             "platform/user-management",
             "platform/organization-auth-tokens",
+            // "platform/device-attributes",
```
sirlantis, line 19:
```suggestion
            "platform/device-attributes",
```

PR #102, opened 2021-09-02
MFLT-4273 Add emb Linux SWUpdate integration guide w sample config

martijnthe:


docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,293 @@
...
+
+### Project Level Filtering
+
+Within a "Project", data can be filtered on three dimensions:
+
+- `Device` - An edge device that communicates with the Memfault cloud. You can
+  look at data collected at the device level to identify the _exact_ issues
+  impacting that unit.
+- `Cohort` - A grouping of devices. You will be able to filter data and deploy
+  releases by this granularity. For example, you could use a cohort to represent
```
fnune, line 34: `by this granularity` isn't strictly true when factoring in primary software types to which an OTA payload can be deployed, or upcoming Delta Releases which can affect Devices in a Cohort of a specific SoftwareVersion.

Nitpicking, though, I think this is good for the purposes of this doc.

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,293 @@
...
+  devices used for different release stages ("alpha", "beta", "prod") or to
+  identify devices by location ("office", "site A", "site B") or a combination
+  of both ("Site A Alpha", "Site A Beta", etc). A device is always assigned to
+  exactly one "Cohort" at any given time but can freely be moved to another at
+  any time.
+- `Fleet` - All devices within a Project
```
fnune, line 40: Wondering why this is the only bullet point in this list without a finishing period. I guess because it's not a full sentence (i.e. it lacks a verb)?

sidebars.js
```diff
@@ -81,6 +81,10 @@ module.exports = {
             "android/android-releases-integration-guide",
             "android/android-data-scrubbing",
         ],
+        "Embedded Linux Guides": [
+            "linux/introduction",
+            "linux/linux-releases-integration-guide",
+        ],
```
fnune, line 87: Comment this out so that it's not visible until later?

docusaurus.config.js
```diff
@@ -45,6 +45,12 @@ module.exports = {
                     label: "MCU",
                     position: "left",
                 },
+                {
+                    to: "docs/linux/introduction",
+                    activeBasePath: "docs/linux",
+                    label: "Embedded Linux",
+                    position: "left",
+                },
```
fnune, line 53: [2] Comment this out so that it's not visible until later?

fnune (approved):
I started with a super careful review but realized it's mostly the Android docs copied over. So feel free to ignore those comments :)

If we comment out the menu items we could merge this and iterate on top of it?

docs/linux/introduction.mdx
```diff
@@ -0,0 +1,21 @@
...
+title: Introduction
+sidebar_label: Introduction
+---
+
+Memfault is a full device monitoring solution which can be integrated into any
+Embedded Linux based device.
+
+Memfault's support for Embedded Linux devices is a work-in-progress.
+
+OTA / Release Management is currently fully supported through an off-the-shelve
```
franc0is, line 12:
```suggestion
OTA / Release Management is currently fully supported through an off-the-shelf
```

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,293 @@
+---
+id: linux-releases-integration-guide
+title: Embedded Linux OTA Integration Guide
+sidebar_label: OTA / Release Management
+---
+
+This tutorial covers how to manage your Android over-the-air update flow with
```
franc0is, line 7:
```suggestion
This tutorial covers how to manage your Linux over-the-air update flow with
```

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,293 @@
...
+- `Cohort` - A grouping of devices. You will be able to filter data and deploy
+  releases by this granularity. For example, you could use a cohort to represent
+  devices used for different release stages ("alpha", "beta", "prod") or to
+  identify devices by location ("office", "site A", "site B") or a combination
+  of both ("Site A Alpha", "Site A Beta", etc). A device is always assigned to
+  exactly one "Cohort" at any given time but can freely be moved to another at
+  any time.
+- `Fleet` - All devices within a Project
+
+### Identifying an Android Device
```
franc0is, line 42:
```suggestion
### Identifying a Linux Device
```

docs/linux/linux-releases-guide.mdx
martijnthe: Thanks, yah this file is just a stub / a verbatim copy of the android one.
We'll need to change all of this.

fnune:
Full-page screenshots. The differences are only in the `Query for OTA Payload` section:

| Selecting the SWUpdate tab | Selecting the "Custom" tab |
| --- | --- |
| image | image |

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,435 @@
...
+import LinuxOtaOptionsTabs, {
+  OtaSwupdateTab,
+  OtaCustomTab,
+} from "@site/src/components/LinuxOtaOptionsTabs";
+import TabItem from "@theme/TabItem";
+
+This tutorial covers how to manage your Embedded Linux over-the-air update flow
+with Memfault!
+
+## Terminology Overview
```
martijnthe, line 16: nit: maybe put in a shared mdx snipped / component?

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,435 @@
...
+);
+```
+
+:::note
+
+When SWUpdate downloads and applies your update, it will _not_ automatically
+update the `$SOFTWARE_VERSION` in this file. It is the responsibility of your
+setup to do this. We recommend generating a UUID during boot time (if it doesn't
+exist already) and storing it in non-volatile storage. You may choose a storage
+location that will survive updates and/or factory resets.
```
martijnthe, line 281: I'm confused. What does the generated UUID have to do with `$SOFTWARE_VERSION`? I thought this was an alterative idea for a device serial.

I think I would just omit the last 2 sentences.
```suggestion
setup to update the `swupdate.cfg` file in the .swu payload to match the new version.
```

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,435 @@
...
+
+:::
+
+#### Example SWUpdate invocation
+
+`swupdate -f /etc/swupdate.cfg -u "--polldelay 3600"`
+
+Assuming you've stored `swupdate.cfg` in `/etc/swupdate.cfg`, this command will
+start SWUpdate in Suricatta daemon mode and configure the daemon to poll
+Memfault servers every 3600 seconds.
```
martijnthe, line 291: Maybe lower to once every 24 hours?

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,435 @@
...
+
```
martijnthe, line 292: Somewhat related: also wonder whether we should show how to add a random delay before starting to avoid "thundering herd" effects (i.e. when a whole fleet reboots at the same time).

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,435 @@
...
+The [`meta-memfault-swupdate-example`][6] Yocto layer is based on [SWUpdate
+documentation on building with Yocto][7] and should serve as an example of
+building your own Linux image that contains SWUpdate preconfigured to work with
+Memfault.
+
+The following `Dockerfile` serves as an example of using
+`meta-memfault-swupdate-example` as part of a Poky build:
+
+```Dockerfile
```
martijnthe, line 303: I would perhaps just link to the file in the repo. This might fall out of sync quickly. I also think having this here makes the page look more complicated than it really is.

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,435 @@
...
+locate this token you will also want to navigate to the "Settings" -> "General"
+page in the Memfault UI and copy/paste the token in the "Project Key" section.
+
+## Managing Your First Release
+
+### Create Release
+
+In the commands below, please replace `${YOUR_SOFTWARE_TYPE}` with the Software
+Type this Release contains. For example, in case of a BeagleBone Black we may
+give a Software Type of `bbb-fw`.
```
martijnthe, line 133: Maybe explain briefly that the .swu file is the OTA payload we're going to upload. And perhaps mention any type of payload is supported but in the example here we're using a SWUpdate payload?

martijnthe (commented):
:heavy_check_mark: (Approve!) not sure how to transfer ownership of the PR to you (and can't approve because I'm the owner)

docs/linux/linux-releases-guide.mdx
fnune: > Maybe explain briefly that the .swu file is the OTA payload we're going to upload

:+1: 

> mention any type of payload is supported but in the example here we're using a SWUpdate payload

:+1:

docs/linux/linux-releases-guide.mdx
fnune: Oops, just got distracted in the middle of writing that and I just made no sense. I guess I could mention generating device IDs in a separate paragraph, but that's probably also super target/project-dependent. I'll take your suggestion.

docs/linux/linux-releases-guide.mdx
fnune: I was thinking about doing this. Other Releases guides have sections that differ only a little bit. Also, we'll need to update them all equally carefully when we start talking about the Deployment matrix here.

docs/linux/linux-releases-guide.mdx
HBehrens: 24 hours is too low considering that some devices can only ever update once a day we could miss the time window this way. I know that this is kind of random, but I'd not suggest anything lower than 5 hours (and 1 hour should be fine considering that we already handle the same amount of data with heartbeats today).
Also note that mid-term we want this value to be configurable at runtime via SDK settings (similar to Bort).

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,435 @@
...
+  in the factory or a chip identifier present in the silicon of the main MCU on
+  your platform. This identifier _must_ be unique for every device in the same
+  Project.
+- `Hardware Version` / `hardware_version` - For a given device, the revision of
+  hardware present. For example, this can be used to represent the different
+  stages of [manufacturing builds](https://mflt.io/hardware-engineer-speak)
+  (`evt`, `dvt`, `pvt`, ...).
+
+_Note: For any given device, neither of these pieces of information should ever
+change._
```
HBehrens, line 64: if we mention GUID-based identifiers above, we should briefly indicate the consequence of a factory reset over here too -> will be treated as a new device.

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,435 @@
...
+### Identifying Firmware Binaries
+
+- `OTA Payload` - This is the file that you deliver to your device when
+  performing an over-the-air update.
+- `Release` - For each "Release" a unique "OTA Payload" can be deployed per
+  "hardware_version". A set of "hardware_version" / "OTA Payload" pairs makes up
+  a "Release".
+- `Deployment` - The act of publishing a "Release" to a "Cohort" generates a
+  "Deployment". Once deployed, any device in the given "Cohort" will be able to
+  query and update to this Release.
```
HBehrens, line 86: This paragraph needs some serious rethinking some time soon including a diagram. At that point latest, it's worth extracting this into its own MDX

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,435 @@
...
+:::
+
+### Deploy Release
+
+Now let's deploy the release to the `default` cohort.
+
+:::note
+
+Any new device seen that has not explicitly been assigned to a custom cohort
+will be part of this cohort. To pre-create devices assigned to a specific cohort
```
HBehrens, line 195: I'd write "[…] will be part of the default cohort." to make the note block work by itself (maybe update the Android docs, too?)

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,435 @@
...
+
+:::
+
+In the UI navigate to "Fleet" -> "Cohorts". Click on the picker under next to
+the `default` cohort under "Target Release", select the Release to deploy and
+you will be prompted with options that can be performed for the release.
+
+### Query for OTA Payload
+
+Now that we have a release uploaded let's walk through the steps to download the
```
HBehrens, line 206: uploaded *and published*

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,435 @@
...
+  https://github.com/memfault/memfault-yocto/tree/main/meta-memfault-swupdate-example
+[7]: https://sbabic.github.io/swupdate/building-with-yocto.html
+
+  </TabItem>
+  <TabItem value="custom">
+
+To receive an OTA Payload, a device will need to pass these REST API query
+parameters:
+
+- `device_serial` - This is used for determining the `cohort` a Device is in.
```
HBehrens, line 366: To answer my question from above myself: I believe the transition to "device identifier" has to wait until we have also prepared places like this endpoint to accept either `device_serial` or `device_identifer`

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,435 @@
...
+The following configuration can be used as a starter. Save it in `swupdate.cfg`
+and remember to replace all the placeholders with valid values for your project
+and device:
+
+```libconfig
+globals:
+{
+        verbose = true;
+        loglevel = 6;
+        syslog = true;
```
HBehrens, line 251: should there be a remark in the config to tone down these values in production?

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,435 @@
...
+                {event = "fail"; format="#4,date,fw"}
+        );
+}
+
+identify : (
+        { name = "mpk"; value = "$YOUR_PROJECT_KEY" },
+        { name = "id"; value = "$DEVICE_SERIAL" },
+        { name = "v"; value = "$SOFTWARE_VERSION" },
+        { name = "st"; value = "$SOFTWARE_TYPE" },
+        { name = "hw"; value = "$HARDWARE_VERSION" }
```
HBehrens, line 271: it's unclear to me here if the actual configuration `swupdate.cfg` will have variables such as `$DEVICE_SERIAL` in them. From what I understand, it does not, so maybe insert the actual values similar to the curl examples?

docs/linux/linux-releases-guide.mdx
HBehrens: Yes, please link

HBehrens (commented):
LGTM! I would probably also link to documentation to produce SWUpdate update images.

docs/linux/linux-releases-guide.mdx
fnune: It's right before it:

```
The following configuration can be used as a starter. Save it in `swupdate.cfg`
and remember to replace all the placeholders with valid values for your project
and device:
```

I think adding the actual values would introduce an additional step to figure this out; for example going from `hw` to (e.g.) `evt`, to noticing that `evt` is a `hardware_version`.

docs/linux/linux-releases-guide.mdx
fnune: Yes, I was tempted to include a diagram but we need to consider all the OTA documentation pages together.

docs/linux/linux-releases-guide.mdx
fnune: Agreed, but I was planning to remove the GUID identifier comment from above.

docs/linux/linux-releases-guide.mdx
```diff
@@ -0,0 +1,435 @@
...
+- `Fleet` - All devices within a Project
+
+### Identifying a Device
+
+For a "Device" in a project, several pieces of identifying information are
+tracked.
+
+- `Device Serial` / `device_serial` - A unique identifier of a device. For
+  example, this may be a serial number written into the EEPROM of your product
+  in the factory or a chip identifier present in the silicon of the main MCU on
```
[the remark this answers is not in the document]
fnune: I think it may be a bit of a tangent from the topic of the doc. An integrating customer could follow this doc successfully regardless of whether they decide to implement update/factory-stable GUIDs for their devices.

docs/linux/introduction.mdx
```diff
@@ -0,0 +1,21 @@
+---
+id: introduction
+title: Introduction
+sidebar_label: Introduction
+---
+
+Memfault is a full device monitoring solution which can be integrated into any
+Embedded Linux based device.
```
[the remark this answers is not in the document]
fnune: It wasn't intentional. Also, I'm now tempted to spell this `embedded Linux-based`, or even `embedded-Linux-based` :grimacing: .

docs/linux/linux-releases-guide.mdx
fnune: Ended up with this:

```
#### Example SWUpdate invocation

`swupdate -f /etc/swupdate.cfg -u "--polldelay $(shuf --input-range=14400-21600 --head-count=1)"`

Assuming you've stored `swupdate.cfg` in `/etc/swupdate.cfg`, this command will
start SWUpdate in Suricatta daemon mode and configure the daemon to poll
Memfault servers every periodically at a random interval of between 14400 and
21600 seconds (4 to 6 hours).
```

docs/linux/introduction.mdx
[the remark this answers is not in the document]
fnune: Went for `embedded Linux-based`,  and `embedded Linux` elsewhere.

docs/linux/linux-releases-guide.mdx
fnune: Updated it for Android and MCU as well.

docs/linux/linux-releases-guide.mdx
fnune: Updated this for all platforms.

docs/linux/linux-releases-guide.mdx
fnune: Added a remark in the introductory list above.

PR #110, opened 2021-09-24
Add note about -flto when uploading symbols

noahp:


noahp:
I ran across this today... not sure it's something a customer encounters but added the note here.

martijnthe (approved):

docs/mcu/symbol-file-build-ids.mdx
```diff
@@ -81,6 +81,33 @@ memfault --org-token $ORG_TOKEN --org acme-inc --project smart-sink \
...
+--software-type options or add a Build Id (see https://mflt.io/symbol-file-build-ids)
+Usage: memfault upload-mcu-symbols [OPTIONS] PATH
+
+Error: Upload failed!
+```
+
+`-flto` can inline the contents of the build id variable, removing the symbol
+from the executable. To work around this, add this flag to the linker args:
+
+`-Wl,--require-defined=g_memfault_build_id`
```
martijnthe, line 105: Would adding `__attribute__((used))` to the variable perhaps also be a solution?

noahp: I thought it would do it, but the linker still removed the symbol reference- the actual data was still present at the correct location:

```
❯ arm-none-eabi-readelf -t build/main.elf | grep -A2 build
  [ 2] .note.gnu.build-id
       NOTE            08006494 016494 000024 00   0   0  4
       [00000002]: ALLOC
```

In the .map file:

```
Discarded input sections
...
 .rodata.g_memfault_build_id
                0x0000000000000000        0xc /tmp/main.elf.roLltr.ltrans0.ltrans.o
```

martijnthe: TIL! thanks

chrisc11 (approved):

PR #117, opened 2021-10-14
Fix some broken links

tyhoff:
Docusaurus gets really confused when trying to find dead links.
It says it uses the ID for these links, but I can't actually
confirm that. Anywho, fix the links that I found that were broken.


fnune (changes requested):
The links aren't broken. Maybe you built with an old version of Docusaurus before running `yarn`?

HBehrens:
Based on comment by @fnune , can we close this PR, @tyhoff ?

fnune:
Closing, the links are fine, e.g. https://docs.memfault.com/docs/android/android-ota-update-client

PR #129, opened 2021-11-17
Add Inter-italic var

noahp:
This fixes elements styled with `italic` on Chrome (Firefox was doing
some fallback so they rendered as italic but not in Inter font AFAIK).

Update the Inter-roman.var archive to the same version, downloaded from
here:

https://github.com/rsms/inter/releases/download/v3.19/Inter-3.19.zip

This fixes italic elements (eg `<emp>` from Markdown `_ _` tags):

image


fnune (approved):
Sweet! Thx!

PR #141, opened 2021-11-26
Document all the AOSP rate limits

martijnthe:
Screenshot from 2021-11-26 09-44-54


fnune (approved):
Sweet! THX

Wondering if it's worth adding matching warning in the linked-to docs. For example, for Metrics, we did this:

```md
Apart from being limited by quotas, the ingestion of Metrics may be
[rate-limited](/docs/platform/rate-limiting#custom-reports). Avoid sending data
more than once per hour per device.
```

martijnthe:
> Wondering if it's worth adding matching warning in the linked-to docs.

Added!

fnune (approved):

vlindhol (approved):
👍

HBehrens:
It's a nit but I wonder of we should aim for a consistent order when we mention,

1. MCU,
2. Android, and 
3. Embedded Linux.

Our docs use this order in our side bar, albeit the top bar navigation diverges (Android last). Your table seems to use alphabetical order (which would lead to Android, Embedded Linux, MCU) but I would guess that you'd have chosen "All" as the first item regardless…

martijnthe:
@HBehrens will change the order to All, MCU, Android, Linux in a separate PR.

PR #155, opened 2021-12-24
Migrate to DocSearch V3

fnune:
https://docusaurus.io/blog/2021/11/21/algolia-docsearch-migration

 I received an email stating that all crawlers are being migrated to a
 new platform by Algolia and that we have until February 2022 to
 migrate. Since the old crawler would stop to work in February 2022, we
 have no choice but to migrate.

 - Benefits: we get a dashboard with analytics for our searches.
 - Drawbacks: we're apparently now under a rate-limited [Free
 Plan](https://go.algolia.com/free-plan-terms) with a limit of 10
 "search units". A "Search unit" is described
 [here](https://support.algolia.com/hc/en-us/articles/4406981856785-What-is-a-usage-unit-).

 It's uncertain how our previous ["for
 open-source"](https://www.algolia.com/for-open-source/) agreement
 translates to this.


chrisc11 (approved):
Thanks @fnune !

PR #160, opened 2022-01-12
Add note about editing metrics for quotas

noahp:
Extracted the screenshot from chrome dev tools "Capture node
screenshot", then added a 1px border with

```bash
❯ convert -bordercolor black -border 1 static/img/docs/platform/metric-edit{,out}.png
```


chrisc11 (approved):
<3

PR #181, opened 2022-02-22
Fix URLs when current page includes trailing `/`

noahp:
@dmyablonski noticed this URL issue; affects any link that isn't either
absolute or relative or includes the `.mdx` trailing file extension.

Install ripgrep with this command to get PCRE2:

```bash
❯ cargo install ripgrep --features 'pcre2'
```

Then use this spicy command to find URLs that likely have the issue
(also uses fd-find):

```bash
❯ rg --engine pcre2 '\]\([^/h#\.](?![^/:]*?\.mdx)[^:/]*?\)' $(fd '\.mdx' | sort)
```

See also
https://github.com/dmyablonski/memfault-docs/commit/01bfe5e2385303be7f056c47e43ddd82144bd64f
.

### DIFF
--- a/docs/android/android-builtin-metrics.mdx
+++ b/docs/android/android-builtin-metrics.mdx
@@ -10,7 +10,8 @@ from Android devices.
 Some are automatically collected, and some require configuration in the Memfault
 dashboard.
 
-To record custom metrics, see the [Reporting APIs](android-custom-metrics).
+To record custom metrics, see the
+[Reporting APIs](/docs/android/android-custom-metrics).
 
 :::info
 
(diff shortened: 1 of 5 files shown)


HBehrens:
I believe @fnune identified something similar recently. Does it make sense to include this as an automation in our gh-action step that today only prettifies?

fnune (approved):
:smile: Thanks for the fix and the regex-fu

noahp:
> I believe @fnune identified something similar recently. Does it make sense to include this as an automation in our gh-action step that today only prettifies?

You read my mind! Got a follow up that adds a couple of sanity checks.

noahp:
Ah, this appears to be the relevant setting fwiw: https://docusaurus.io/docs/api/docusaurus-config#trailing-slash
haven't really dug into it.

PR #182, opened 2022-02-23
Add a small script for checking page ids

noahp:
This script checks if a page `id: ` tag matches the basename (without
extension) of the file, and returns non-zero if there's a mismatch.

This helps reduce confusion IMO, and I didn't see a reason to not
require it.

Add the check to github actions.

Correcting the ones that don't match by following this strategy:

- rename the _file_ (NOT the ID) to match the ID. this preserves
  external links, since the slug is based on the ID. the file name is
  not used in the generated site.
- update any cross links here that use the file name to match the new
  name (docusaurus fails to compile when they're broken 😍)
- run prettier (npx prettier --write $(gd HEAD --name-only))


fnune (approved):
Cool!

Nit: you included some instructions in the PR description. But when this command fails, the user will only see the mismatches on CI.

Should we include those instructions in the CLI output when it fails? Also maybe the reason why the IDs need to match?

noahp:
> Cool!
> 
> Nit: you included some instructions in the PR description. But when this command fails, the user will only see the mismatches on CI.
> 
> Should we include those instructions in the CLI output when it fails? Also maybe the reason why the IDs need to match?

Good call, I'll plop it in there!

PR #183, opened 2022-02-23
Update docusaurus to 2.0.0-beta.15

noahp:
A lot of changes since `-beta.0`. I'm pulling this in to resolve the
dependabot auto-PRs en masse.

Updated our TargetVersionFilled darkmode sensitive function with the API
changes in docusaurus.

Add a `.node-version` file indicating we need ^14 now.

Also update the .forestry image to a node v14 one (I don't know what
that's used for but 🤷)


docusaurus.config.js
```diff
@@ -175,6 +171,10 @@ module.exports = {
                     editUrl:
                         "https://github.com/memfault/memfault-docs/edit/main/",
                 },
+                gtag: {
+                    trackingID: "G-R5JYJ06TDJ",
```
noahp, line 175: TODO confirm that this ends up in the generated site!!!

noahp: Confirmed (I just ran `yarn build && rg --no-ignore -t html G-R5JYJ06TDJ`)

noahp:
@fnune should that `.node-version` file be enough to convince netlify to use nodejs 14? seems to not be picking it up and I don't have access to the site settings :confused: https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript

vlindhol (approved):
LGTM!

chrisc11 (approved):
Cool, I like some of the small style changes in the new version

> Also update the .forestry image to a node v14 one (I don't know what
that's used for but 🤷)

no need to do it in this PR but we aren't using it anymore so can be dropped fwiw

noahp:
Did a pass through most of the pages, didn't see anything suspicious. Merging and let's keep an eye out for a revert.

PR #189, opened 2022-03-10
MFLT-5656 Update documentation on device search

fnune:
The only content that mentions device search is the metrics docs. I
updated the screenshots and the text.

I considered updating Device Sets as well, but the text in that is
generic enough that I felt no update was needed.


martijnthe (approved):
nit: some of the screenshots are pretty wide

fnune:
They take the full width of the text container in the docs... I think they look appropriate there:

image

fnune:
Actually, that first pic I took was of the old docs. The new ones do look a bit squeezed in. Changing them.

PR #192, opened 2022-03-23
fix missing link to device info component

HBehrens:
With 36b9e6cc26df3e2e50931467dce6a60cea53ba7e, we fixed one link to the MCU device info component but missed another. This completes the change.


HBehrens:
Based on [the internal conversation on Slack](https://memfault.slack.com/archives/C019Z4X8SBH/p1647957779323119?thread_ts=1647905715.695159&cid=C019Z4X8SBH) @franc0is editing the link to "the wrong readme". I am still not 100% sure what we should link to here. The new destination is not a README and there is no README closer to the device info component than what we pointed to before.

I wonder if we should revert the links back to the README (they are now pointing at the header) and improve the README instead. @chrisc11 ?

PR #199, opened 2022-04-12
Thread analysis initial doc

martijnthe:
Added a page on RTOS analysis, documenting what needs to be captured
in order to get threads properly analyzed. I focused on FreeRTOS and Zephyr for now,
hoping to add more detail for others some time in the future, as we see fit.

PR #201, opened 2022-04-12
Add some tips about FreeRTOS task decode

noahp:


noahp:
Oh wow, completely missed #199, which does a better job of documenting this!

PR #203, opened 2022-04-13
Fix a bug showing the right logo after SSR

fnune:
Looks like the logo was defaulting to the light variant of `src` during
SSR, and it didn't get updated once React rehydrated the component.
Adding `isClient` as a key should prompt a rerender with the correct
theme.

https://github.com/facebook/docusaurus/issues/2912


fnune:
YOLO-merging. No need to bother anyone with this.

PR #210, opened 2022-04-15
Add note about running gdb-multiarch in docker

noahp:


docs/mcu/coredump-elf-with-gdb.mdx
```diff
@@ -151,3 +151,26 @@ For help, type "help".
...
+
+```bash
+# the first line below is the docker run command. the second line is the gdb
+# command running from the current working directory, which is mounted inside
+# the container ('--volume' argument)
+$ docker run --rm --tty --interactive --volume "$PWD":/data memfault:multi-arch-gdb \
+  gdb --se example/symbol_file.elf --core example/coredump.elf
+```
+
+Example usage:
```
noahp, line 174: May want to omit the example image, could be confusing

chrisc11: I like the image

chrisc11 (approved):

PR #211, opened 2022-04-21
mcu/compact_logs: explicit `__start_log_fmt` def

noahp:
Some older versions of binutils don't seem to reliably place the
`__start_SECTION` automatic symbol, so include an explicit definition so
our compact log decoding will work. An example of a toolchain with the
problem is this one:

https://github.com/espressif/esp-idf/blob/v4.2.3/tools/tools.json#L30

Update our copy of `prismjs` to support _cutting edge_ GNU linker-script
syntax highlighting:

https://github.com/PrismJS/prism/pull/3373

### DIFF
--- a/docs/mcu/compact-logs.mdx
+++ b/docs/mcu/compact-logs.mdx
@@ -31,7 +31,7 @@ MEMFAULT_TRACE_EVENT_WITH_LOG(flash_error, "QSPI Flash Erase Failure: error code
 With a formatted string we need to encode the string length as well as the
 entire string in our binary message taking up 41 bytes
 
-```
+```plaintext
 00000000: 5153 5049 2046 6c61 7368 2045 7261 7365  QSPI Flash Erase
 00000010: 2046 6169 6c75 7265 3a20 6572 726f 7220   Failure: error
 00000020: 636f 6465 3a20 3078 35                   code: 0x5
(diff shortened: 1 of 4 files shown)


docs/mcu/compact-logs.mdx
```diff
@@ -63,13 +63,17 @@ Add the following to your projects `memfault_platform_config.h`:
...
-```
+```linker-script
     log_fmt 0xF0000000 (INFO):
     {
-       KEEP(*(*.log_fmt_hdr))
-       KEEP(*(log_fmt))
+        /*
+        some versions of binutils do not place the __start_SECTION automatic
+        section symbol correctly by default, so explicitly add it here
+        */
```
chrisc11, line 72: Maybe something like:

/* Note: binutils >= 0.29 will automatically create this symbol but we set explicitly to support older versions */

chrisc11 (approved):
thanks!

docusaurus.config.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
         },
         prism: {
             theme: require("prism-react-renderer/themes/oceanicNext"),
+            additionalLanguages: ["linker-script"],
```
fnune, line 25: Cool!

fnune (approved):

PR #216, opened 2022-04-26
Document our CDN's features in the OTA doc

fnune:
Also add the same admonition to the top of each platform-specific OTA
guide.

image


franc0is (approved):
👍  you might remove "edge rules" because that's not relevant to the customer

fnune:
Perfect TY

PR #218, opened 2022-04-26
Docs: add hawkBit comparison page

topher200:
### Summary of changes
Create a page for comparing hawkBit OTA to Memfault OTA


topher200:
* main
  * **PR #217** <a href="https://app.graphite.dev/github/pr/memfault/memfault-docs/217" target="_blank">Graphite</a> 
    * **PR #218** <a href="https://app.graphite.dev/github/pr/memfault/memfault-docs/218" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/memfault/memfault-docs/218?utm_source=stack-comment).

docs/linux/linux-memfault-hawkbit-comparison.mdx
```diff
@@ -0,0 +1,26 @@
...
+title: Memfault OTA for hawkBit users
+sidebar_label: Memfault OTA for hawkBit users
+---
+
+Memfault's Embedded Linux OTA solution is hawkBit API compatable. For
+information on Linux OTA, please see
+[Embedded Linux OTA Integration Guide](/docs/linux/linux-releases-integration-guide).
+
+This page is aimed at users who are already familiar with the hawkBit API and
+UI. We demonstrate the differences between Memfault and hawkBit.
```
HBehrens, line 12: can we provide links here (in addition to or instead of the reference below)?

docs/linux/linux-memfault-hawkbit-comparison.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+| hawkBit             | Memfault      | What it's used for                                                                                                                                                                                                                         |
+| ------------------- | ------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| Tenant              | Project ID    | Namespacing groups of devices.                                                                                                                                                                                                             |
```
HBehrens, line 16: how does that relate to the structure of our DDI API where we ignore the tenant ID? And why do we call it "tenant" without "ID" but "project ID" *with* "ID"?

docs/linux/linux-memfault-hawkbit-comparison.mdx
```diff
@@ -0,0 +1,26 @@
...
+| Provisioning Target | Device        | System to update.                                                                                                                                                                                                                          |
```
HBehrens, line 17: for the different Memfault entities, does it make sense to link to the respective Memfault docs? After call, the goal if this page is for hawkBit users to familiarize themselves with Memfault.

docs/linux/linux-memfault-hawkbit-comparison.mdx
```diff
@@ -0,0 +1,26 @@
...
+| Controller ID       | Device serial | Identifying specific devices. This value must be unique per Project.                                                                                                                                                                       |
```
HBehrens, line 18: we try to steer away from the term "device serial" in favor of "device id"

docs/linux/linux-memfault-hawkbit-comparison.mdx
```diff
@@ -0,0 +1,26 @@
...
+| Artifact            | OTA Payload   | Uploaded binary file. A Memfault OTA Payload targets a specific Software Type.                                                                                                                                                             |
```
HBehrens, line 19: it targets a specific hardware version, which you have to choose when uploading payload. Different hard version versions may all refer to the same software type.

docs/linux/linux-memfault-hawkbit-comparison.mdx
```diff
@@ -0,0 +1,26 @@
...
+| Software Module     | OTA Payload   | A hawkBit Software Module is a versioned collection of Artifacts. A Software Module has an assignment (firmware/OS or software/Application). Memfault OTA Payloads include a Software Type, removing the need to group by Software Module. |
+| Distribution        | Release       | A hawkBit Distribution is a versioned set of Software Modules. A Memfault Release is a versioned set of OTA Payloads.                                                                                                                      |
```
HBehrens, line 21: If the assumption is that readers are familiar with hawkBit concepts, why do we need to explain them here? Does it make more sense to focus on the differences only? e.g. why does the Software Type remove the need to group?

HBehrens (commented):
I think it is worthwhile to explain the bigger picture on this page:

1. hawkBit Clients on the device remain unchanged but need to be configured to point to Memfault since
2. Memfault acts as a hawkBit update server that implements hawkBit DDI.
3. Users leverage the Memfault Web Application as a replacement of the hawkBit Management UI.

Also worth calling out: Memfault does not implement the hawkBit Management API – instead, for headless OTA management, one would need to exercise the Memfault API.

docs/linux/linux-memfault-hawkbit-comparison.mdx
topher200: > how does that relate to the structure of our DDI API where we ignore the tenant ID? 

Do we ignore the tenant ID? The way I understand it, where the DDI API docs say to use `tenant_id` we expect that the Memfault customer will instead use their Memfault `project_id`. @vlindhol Is it accurate to say that anywhere one sees `tenant_id` in the DDI docs they should expect to instead use their Memfault `project_id`?

Please advise if I'm misunderstanding your question! But `project_id` seems like a perfect analogy to `tenant_id` to me. If I'm misunderstanding, please advise what text I should use instead.

> And why do we call it "tenant" without "ID" but "project ID" with "ID"?

Good call, I will unify them.

docs/linux/linux-memfault-hawkbit-comparison.mdx
topher200: Great callout!

docs/linux/linux-memfault-hawkbit-comparison.mdx
topher200: Some of the stated goals of these docs (from what I understand):

- (tactical) help existing hawkBit users who are trying to implement Memfault
- (marketing) explain the differences between the two (and show that we have feature parity)
- (internal) explain to Memfault employees the differences between the two platforms so they can speak intelligently about them, as well as develop new features for both

I was hoping to accomplish all three of these goals with these docs! it's possible that we need to make a separate Notion page for that last one (but I'm hoping to avoid that).

Anyway, that's why I included explanations of how hawkBit works. It felt like it would be useful for Memfault-ers trying to understand hawkBit, and for existing hawkBit users to understand the analogues. (even if you're an existing hawkBit user, that doesn't mean you're an expert in all the stated concepts/nouns!).

Perhaps I should delete the explanations and move them to Notion?

---

> why does the Software Type remove the need to group?

I'm not sure if it does! Can you think of a better explanation for why Memfault doesn't have the concept of "Software Modules"?

docs/linux/linux-memfault-hawkbit-comparison.mdx
topher200: I'm not quite sure I understand, just looking at the product. It looks like you need to specify a software version when creating an OTA Payload. Hardware versions aren't mentioned.


https://user-images.githubusercontent.com/206988/165633672-def6e776-689a-4841-a3a1-55e124c96549.mp4

docs/linux/linux-memfault-hawkbit-comparison.mdx
topher200: Topher will remove hawkBit explanations and include links instead

docs/linux/linux-memfault-hawkbit-comparison.mdx
topher200: We just use `default`. I'm going to remove the 'tenant' row completely.

fnune (approved):
:+1: There's a bit of bad syntax in `hawkBit[5]` but it results in a usable link, so approving. I think you meant `[hawkBit][5]`.

**EDIT:** I meant to leave this review in the other PR... https://github.com/memfault/memfault-docs/pull/217

docs/linux/linux-memfault-hawkbit-comparison.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+## Reference
+
+- [Memfault Terminology](/docs/platform/memfault-terminology)
+
+[hawkbit-ddi-api]: https://www.eclipse.org/hawkbit/apis/ddi_api
+[hawkbit-management-ui]: https://www.eclipse.org/hawkbit/ui
+[hawkbit-management-api]: https://www.eclipse.org/hawkbit/apis/management_api
+[memfault-release-management-ui]: /docs/platform/ota
+[memfault-linux-ota-guide]: /docs/linux/linux-releases-integration-guide
```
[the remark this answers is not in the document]
fnune: https://github.com/memfault/memfault-docs/pull/217#discussion_r861788272

The problem is that either Prettier or Pycharm broke clusters of links like this one by introducing line breaks in the wrong places. If this is an editor problem then we should still adopt them!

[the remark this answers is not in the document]
topher200: i'll keep using them when the URL is short (no formatting issues!) and avoid them when not.

it also looks like it's possible to block `prettier` for a block: https://prettier.io/docs/en/ignore.html

docs/linux/linux-memfault-hawkbit-comparison.mdx
```diff
@@ -0,0 +1,48 @@
...
+For general information about Memfault OTA, including all the features
+available, please see our [OTA Update docs][memfault-release-management-ui].
+
+This page is aimed at users who are already familiar with the [hawkBit DDI
+API][hawkbit-ddi-api] and [Management UI][hawkbit-management-ui].
+
+## Differences between the platforms
+
+For migrations to Memfault, hawkBit Clients on the device remain unchanged but
+need to be configured to point to Memfault;
```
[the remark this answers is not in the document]
topher200: I added a comment callout out requiring Gateway auth

docs/linux/linux-memfault-hawkbit-comparison.mdx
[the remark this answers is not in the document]
topher200: The prettier issue isn't happening for me locally; the links are wrapping to a new line but docusaurus is handling them fine.

I'll confirm it's working correctly after i push to prod

PR #224, opened 2022-05-09
Add example snippet for string metrics

noahp:


docs/mcu/metrics-api.mdx
```diff
@@ -96,6 +97,24 @@ void bluetooth_driver_send_bytes(const void *data, size_t data_len) {
...
+                                        manufacturing_date);
+
+  // optionally, trigger a heartbeat to immediately capture the metric record
+  memfault_metrics_heartbeat_debug_trigger();
+
+  // optionally, trigger an upload of Memfault chunk data
+  // [ ... code to trigger memfault upload ... ]
+}
+```
+
```
chrisc11, line 117: Maybe a callout -- if a string metric is not reported in a heartbeat interval, the previously reported value will not be overwritten by Memfault's backend. This can be used as a bandwidth optimization by only reporting values on bootup or when they change.

chrisc11 (approved):
ty!

docs/mcu/metrics-api.mdx
noahp: Oh good call, will add

PR #226, opened 2022-05-12
Compare Memfault and hawkBit Management UI

victorbordo:
### Summary
Update the hawkBit docs with a visual comparison between Memfault
and the hawkBit Management UI.

PR #234, opened 2022-05-24
Update hawkBit UI comparison

victorbordo:
### Summary
New version to address feedback from #226.


docs/linux/linux-memfault-hawkbit-comparison.mdx
```diff
@@ -45,28 +45,16 @@ Here is an overview of hawkBit concepts and how they map to Memfault:
...
-> - Deployment Management view for target administration and manual deployment.
-> - Distribution Management view software repository metadata management.
-> - Artifact Management view to manage the artifacts.
-> - Target Filter Management view to manage target filters that can be used both
->   in Deployment and Rollout Management views.
-> - Rollout Management for large scale rollout orchestration.
+Let's compare the [hawkBit management UI](https://www.eclipse.org/hawkbit/ui/)
+and Memfault for
+[managing OTA updates](/docs/platform/ota/#managing-ota-updates-with-memfault).
+The goal of this section is to highlight the differences between hawkBit's
```
HBehrens, line 51: is this really the goal and what we're doing here? Don't we rather explain how one would transition from hawkBit to Memfault (i.e. we don't explain how one would map staged rollouts back to hawkBit)?

docs/linux/linux-memfault-hawkbit-comparison.mdx
```diff
@@ -111,53 +99,19 @@ single view.
...
-    alt="/img/docs/linux/memfault-devices.png"
-  />
-</p>
-
-We can do better. Devices are managed in **Cohorts**, which are logical groups
-of devices. Cohorts are similar to hawkBit Target Filters. Navigate to
-`Fleet/Cohorts`. A `default` cohort is available for every new project. You can
-also create your own cohorts.
+Memfault displays device and release data differently. Navigate to
+`Fleet/Devices` to see your devices. You can learn more about
```
HBehrens, line 103: We can now deep-link to paths like these. As this is brand new, there's no precedence for it (in particular now styling you could copy this from) but the corresponding placeholder link here would be https://app.memfault.com/organizations/-/projects/-/devices

Same for other places (e.g. the releases below)

HBehrens (approved):
Some minor remarks (finally a chance to apply our placeholder links!). Otherwise, LGMT!

I am expecting to iterate on this once we get feedback from customers who integrate and ask specific hawkBit-oriented questions. Until then, let's leave it as-is.

PR #237, opened 2022-05-26
Change mcuxpresso docs link to include nxp

victorbordo:
### Summary
Update link for potential search improvement.


noahp (approved):
:+1:

fnune (approved):
LGTM. Do we need to add something to [`_redirects`](https://github.com/memfault/memfault-docs/blob/main/static/_redirects)?

E.g.

```
# Redirect from the old ARM MCUXpresso guide
/docs/mcu/arm-mcuxpresso-guide /docs/mcu/arm-nxp-mcuxpresso-guide
```

One note though: if the purpose of this is to match this article when users search for `NXP` then we'll need to change the title, too? Currently it doesn't have `NXP` in it. Suggestion: `NXP MCUXpresso Getting Started Guide for ARM`.

victorbordo:
> LGTM. Do we need to add something to [`_redirects`](https://github.com/memfault/memfault-docs/blob/main/static/_redirects)?

I think we can get away without a redirect here due to these docs being so new.

PR #239, opened 2022-06-02
Remove yocto layer link

topher200:
### Issues addressed
 The old yocto layer is no longer relevant, but the new one isn't ready
 yet.

  ### Summary of changes
  Remove the link. Customers will have to generate their own yocto layer.


HBehrens:
I agree that we shouldn't link to an outdated example. If customers have to create their own Yocto now, should we at least say so instead of only removing the paragraphs, @topher200 ?

topher200:
@HBehrens Done!

topher200 (commented):
@victorbordo May I merge this now?

victorbordo (approved):
merge away

PR #241, opened 2022-06-06
Add a note about non-volatile event storage

noahp:


docs/mcu/metrics-api.mdx
```diff
@@ -161,3 +161,44 @@ are a few easy ways you can debug and test the new code. Notably:
...
+  memfault_trace_event_boot(evt_storage);
+  // record the current reboot reason
+  memfault_reboot_tracking_collect_reset_info(evt_storage);
+
+  // configure the metrics component to store into the buffer
+  sMemfaultMetricBootInfo boot_info = {
+    .unexpected_reboot_count = memfault_reboot_tracking_get_crash_count(),
+  };
+  memfault_metrics_boot(evt_storage, &boot_info);
+```
```
HBehrens, line 188: I think there are other places about this, too – any idea how we will know when this copy goes out of sync with the linked source at head of main? Even though we're linking to a specific revision, future readers may get confused about it not matching their more recent SDK.

HBehrens (approved):
LGTM, one non-critical question + nit

docs/mcu/metrics-api.mdx
noahp: hmm, i can add a disclaimer to this snippet, probably the simplest option

ideally i think anytime we reference code, we would have something like a rust-playground/jupyter-like embed where the snippets are run as part of our CI and kept up-to-date... way OT but i really want a way for a user to Try Memfault in-browser via some qemu shell, where they can directly upload from the emulated device

docs/mcu/metrics-api.mdx
```diff
@@ -161,3 +161,43 @@ are a few easy ways you can debug and test the new code. Notably:
 - The heartbeat interval can be reduced from the default 3600 seconds for debug
   purposes by setting `MEMFAULT_METRICS_HEARTBEAT_INTERVAL_SECS` in your
   `memfault_platform_config.h` interval to a shorter period such as 30 seconds.
+
+## Metrics Storage
+
+Metric events are stored in the in-memory ring buffer supplied to the
+`memfault_metrics_boot()` initialization function (snippet below is from the
+[`ports/templates`](https://github.com/memfault/memfault-firmware-sdk/blob/bdacc2d187c59ee3f8f7d262df3e68e8aaeb1b6a/ports/templates/memfault_platform_port.c#L108-L118)
+example, with extra annotations):
```
chrisc11, line 170: OT: I'd be in favor of pulling annotations into template

noahp: https://github.com/memfault/memfault/pull/6776

chrisc11 (approved):

PR #255, opened 2022-06-24
Add “Ideal Image” plugin

lunelson:
This PR adds the @docusaurus-ideal-image plugin, to optimize image sizes and cumulative layout shift


fnune (approved):
Docs for other readers: https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-ideal-image

Love:

- The fact that this automates image optimization and we can now remove my hacky scripts
- That this solves layout shift :+1: 

Not a huge fan:

- The blurred look of images before the user scrolls down to them. I think for our type of users, where they're probably going to be on their workstations reading our docs, we should just load the full-sized images by default. The blur -> hifi shift is a bit distracting to me.

I've used the blur -> hifi technique for other sites but they were consumer websites with long scrolling and so on...

WDYT?

lunelson:
@fnune I'll take a close look, as to whether this LQIP technique can be overridden, but the principle of it is of course that there's some kind of placeholder before the image loads... so an option would just be a solid color, for example, wdyt?

fnune:
May look better yeah! I think the blur doesn't look amazing because it doesn't work super well on straight lines and right angles (IMO) which is what most of our screenshots are.

lunelson:
@fnune unfortunately [although the underlying library does accept a placeholder prop, which will accept a color value](https://github.com/endiliey/react-ideal-image#placeholder), the Docusaurus plugin's exported component doesn't expose it 😐

lunelson:
@fnune although I think the blur doesn't look too bad and we can mitigate the problem by adding a faint outset outline on all images, which will frame them up and make them look cleaner I think.

Also discovered: testing on throttled network exposes a feature, which is that images are not downloaded but the user is presented with a 'click to download' button, pretty nice

noahp:
Ooo, love this idea! Would definitely be great to remove the manual stuff (though :heart: Fausto for doing that, really made a big difference to load speed).

> we should just load the full-sized images by default

+1, I would very much like to keep this behavior if possible!

Also I noticed there seems to be duplicate network requests with this change (might not be real? i don't know anything about web :frowning: ):

Current deploy:
image


This patch:
image

lunelson:
@fnune @noahp would love to get some more feedback on this. I vendorized Docusaurus's component here so we could get solid color fallbacks instead of blurhash, but I'm still seeing the apparent double request that Noah pointed out 🤔

src/theme/IdealImage2.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import React from "react";
+import IdealImage from "@endiliey/react-ideal-image";
+
+function Image(props) {
```
fnune, line 10: Can this file be removed?

src/theme/IdealImage.js
```diff
@@ -0,0 +1,119 @@
...
+import ReactIdealImage from "@endiliey/react-ideal-image";
+import { translate } from "@docusaurus/Translate";
+import cx from "classnames";
+
+import css from "./IdealImage.module.css";
+
+export default function IdealImage(props) {
+    const { alt, className, img, style, caption } = props;
+    // In dev env just use regular img with original file
+    if (typeof img === "string" || "default" in img) {
```
fnune, line 12: Can we use the same implementation we use in production? Debugging will become easier. Also, in this case, how does `typeof img === "string" || "default" in img` translate to being in a dev environment?

fnune (changes requested):
Hmm I'm also seeing the double requests, even after a refresh:

image

Looks like one is initiated by `loaders.js` and the other by `unfetch.js`. Unfetch is a `fetch` polyfill, so it's strange that it's included. But that's probably not the problem.

If I run a production build (`yarn build && yarn serve`) it still happens:

image

The behavior after I keep refreshing is that responses are cached by the browser. It looks like it goes like this:

- First request, image is fetched twice
- Second request, one of requests is downloaded fully again, but the other is cached
- Third request on: all requests are cached

Requesting changes because I believe the double-fetching problem is a regression.

src/theme/IdealImage.js
lunelson: I'm not really sure. I copied this from Docusaurus' plugin directly, so I could override the `placeholder` prop

```
facebook/docusaurus/packages/docusaurus-plugin-ideal-image/src/theme/IdealImage/index.tsx
```

lunelson:
@fnune I'm also thinking about changing the component design, it might be better of captioning just happened through `children` rather than a prop, so it can also contain arbitrary HTML, and the component be called `ImageFigure` as before, for clarity, e.g.:

```jsx
<ImageFigure
	img={debugRegLocal}
	alt="MCU trace analysis view, viewing registers and locals"
>
	<figcaption>
		Our trace analysis view offers the highest resolution available into the crash dumps your devices upload. In this example, you can see the values of local variables and registers at the time of the crash.
	</figcaption>
</ImageFigure>

src/theme/IdealImage.js
fnune: Hmm strange. The comment introduces more questions than answers IMO. ("in dev env...")

fnune:
The first screenshot in my comment is with `yarn start` and the other with a production build with `yarn build && yarn serve`.

fnune:
The behavior on the Netlify preview is the same for me:

image

fnune:
Huh! It seems like it works fine on Chrome, but not on Firefox. See Chrome:

image

The second response is properly cached.

PR #258, opened 2022-06-28
Add a table with links to example projects

noahp:


chrisc11:
I like the idea. I'd probably put at the end of the guide as I could see it getting long and might put board picture in it's own column and shrink the images

fnune (approved):
I think this looks great!

Can we also link to those from the specific guides? Or are they already linked to?

image

PR #261, opened 2022-07-06
MFLT-6497 Optimize ImageFigure component to show original

lunelson:
This extends the range of optimized image sizes to a max of 1920px wide, and wraps the image element in a link to the largest image


HBehrens:
For large images, this seems to work well for me :)

For smaller ones, I wonder what the advice would be. Will we simply stick to `` or can we edit the component to only show the content at 100% in that case? What we have now, upscales the images and makes them appear blurry.

https://docs.memfault.com/docs/platform/organization-auth-tokens#creation
Screen Shot 2022-07-06 at 12 20 42

noahp:
I like this a lot! also quite nice on mobile :+1:

lunelson:
@HBehrens good point, I didn't anticipate that. Sizing is now modified to be "intrinsic" which would not allow images to grow beyond their original size, which avoids that particular bluriness

However this doesn't necessary display images at their "real world" size, for example if they were originally taken on a retina screen

HBehrens:
I believe I’ve seen that though 🙈

lunelson:
> I believe I’ve seen that though 🙈

🤔 not sure what you mean @HBehrens — do you have a further change suggestion, or is this OK?

src/theme/ImageFigure.js
```diff
@@ -13,57 +13,69 @@ export default function ImageFigure(props) {
     const image = useMemo(() => {
         if (typeof data === "string")
```
fnune, line 14: OT: is this related to dev-mode bundling? Is a comment warranted?

fnune (approved):
Looks good to me! Initially I was on the fence about defaulting to `_blank` but I guess it is the lower-friction default.

HBehrens:
> 🤔 not sure what you mean

i believe I saw a @2x retina screenshot that shouldn’t fill the entire width on our repo.

even if not: what’s the proposed workflow for people working exclusively on retina displays when they want to capture and show a screenshot with say 300px logical width? Will we simply not use this new component?

lunelson:
> what’s the proposed workflow for people working exclusively on retina displays when they want to capture and show a screenshot with say 300px logical width? Will we simply not use this new component?

@HBehrens I can expose a `width` prop on the component, that would allow one to specify the intrinsic width, for the case where the pixel density is higher... or alternatively a `pixelRatio` prop which defaults to `1` but would just do the right thing when given `2` for example and not require the author to actually know the precise width, and in any case only be require for images whose intrinsic width is less than about ~900px, which is the rough max width of the column. For the full-screen view opened in a new tab, there's no control as it's a direct link to the image source, not an `img` element

lunelson:
Going for the `pixelRatio` prop option

HBehrens:
Thank you, @lunelson – I think the `pixelRatio` option is what we want here.

For posterity, here would be an image that's captured as @2x https://docs.memfault.com/img/docs/platform/invite-user-min.png – linked from https://docs.memfault.com/docs/platform/user-management

PR #262, opened 2022-07-06
Release Notes for June 2022

HBehrens:


can-memfault:
@HBehrens when I click on "Read more" it scrolls to the beginning of the page. Is that expected? 
Otherwise I checked everything on [here](https://3000-memfault-memfaultdocs-s73d1j97q4e.ws-eu51.gitpod.io/changelog/2022/06/30/) and it looks good!

can-memfault (approved):

HBehrens:
> when I click on "Read more" it scrolls to the beginning of the page. Is that expected?

It never struck me as unusual, but you're right, [Docusaurus' `<truncate>`](https://docusaurus.io/docs/blog#blog-list) could instead link to the first lines that were previously inaccessible instead. Maybe something for you to investigate?

> Otherwise I checked everything on [here](https://3000-memfault-memfaultdocs-s73d1j97q4e.ws-eu51.gitpod.io/changelog/2022/06/30/) and it looks good!

I am glad to see that you're using Gitpod. Please be aware that a) it's exposed ports are by default private. Other people won't have access to the URL you shared and b) workspaces are meant to be short-lived. If you close the tab, the instance that serves the port/URL won't be running a few minutes later.

PR #264, opened 2022-07-20
Improve docs around alert notif frequency

victorbordo:
### Summary
Clarify expectations around when device threshold alert
notifications get delivered.


tyhoff (approved):
Cool, updated the `:::` to be on a new line because it needs to be. Otherwise looks good.

PR #267, opened 2022-07-22
Note about `MEMFAULT_AUTOMATIC_INIT`

noahp:


victorbordo (approved):

docs/mcu/esp32-guide.mdx
```diff
@@ -323,3 +323,29 @@ int memfault_post_data(void) {
...
 ```
+
+## Frequently Asked Questions
+
+### Memfault library initialization
+
+By default, the Memfault library is initialized as part of the ESP-IDF global
+ctors initialization process.
+
+In certain configurations, this can be problematic- for example, if the flash
```
tyhoff, line 334:
```suggestion
In certain configurations, this can be problematic. For example, if the flash
```

tyhoff (approved):
The scary comment makes me question what should be the default :/. 

I'm sure it's a rare case and if it's going to bootloop, it's going to be pretty obvious when it happens.

docs/mcu/esp32-guide.mdx
```diff
@@ -323,3 +323,29 @@ int memfault_post_data(void) {
   return rv;
 }
 ```
+
+## Frequently Asked Questions
+
+### Memfault library initialization
+
+By default, the Memfault library is initialized as part of the ESP-IDF global
+ctors initialization process.
```
chrisc11, line 332: Maybe just remove 334-339.5 and say

```
However, if you would like to defer initialization, the Kconfig flag `CONFIG_MEMFAULT_AUTOMATIC_INIT`
can be set explicitly to `=n`, and then the Memfault library can be manually
initialized by calling the `memfault_boot()` function.
```

noahp: much better, reworked to this :+1:

PR #272, opened 2022-07-26
MFLT-6637 Update Linux SDK instructions to include SSL

topher200:
Remove instructions to unset SSL

 ### Issues addressed
These instructions were required before we had SHA hashes for Linux OTA
artifacts. However, we recently added SHA hashes for artifacts and these
instructions are not required.

 ### Summary of changes
Remove the disable SSL instructions.

 ### Test plan
Followed this advice in the QEMU environment and confirmed I could still
do an update.

### DIFF
--- a/docs/android/android-getting-started-guide.mdx
+++ b/docs/android/android-getting-started-guide.mdx
@@ -220,8 +220,9 @@ that you wish to validate. Connect that device via ADB (verify via
 
 :::info
 
-We recommend running the validation tool with a `userdebug` system image. A `user` image does not allow
-all checks to be run, which may result in Bort configuration issues being missed.
+We recommend running the validation tool with a `userdebug` system image. A
+`user` image does not allow all checks to be run, which may result in Bort
+configuration issues being missed.
 
 :::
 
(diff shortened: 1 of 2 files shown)


topher200:
* main
  * **PR #272** <a href="https://app.graphite.dev/github/pr/memfault/memfault-docs/272" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/memfault/memfault-docs/272?utm_source=stack-comment).

topher200 (commented):
@fnune I found this branch locally from a few months ago. I tested it then but never pushed! I was waiting for SSL support to be completely rolled out in app.memfault.com, which took a fair bit.

I'm surprised that this is all that's required, but then I remembered that the defaults here are OK. `CONFIG_SURICATTA_SSL` is on by default, so there's no need for us to specify it (along with the other ~100 default config vars we don't specify)

fnune:
OK fantastic! Actually great news if we need to document less stuff.

fnune (approved):

PR #276, opened 2022-08-02
MFLT-6592 Document hiding of Issues with zero recent Traces

SionAbes:
MFLT-6592 Document hiding of Issues with zero recent Trace
 
Adds information on why we hide issues with zero traces.

Signed-off-by: SionAbes <sion.n.abraham@gmail.com>

### DIFF
--- a/docs/platform/issue-management.mdx
+++ b/docs/platform/issue-management.mdx
@@ -84,6 +84,5 @@ visualization charts present in the **Issue details** view:
 
 ## Hiding of Issues with zero Traces
 
-As traces are deleted after 60 days Memfault will hide the related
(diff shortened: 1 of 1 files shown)


can-memfault:
@SionAbes I guess your commits are not squashed yet :) 

@HBehrens do you think we need further explanation here? It's just an intermediate step to document current behavior and hopefully we'll have a better way of handling this with minimal traces.

SionAbes:
@can-memfault They get squashed on the merge.. no need to do it in the PR itself.

HBehrens (approved):
Minor change in wording to emphasize that this is happening automatically. Plus one suggestion on punctuation.

Otherwise, LGTM

PR #277, opened 2022-08-02
Add note about updating Memfault SDK in NCS

noahp:


docs/mcu/nrf-connect-sdk-guide.mdx
```diff
@@ -586,3 +586,72 @@ See the list in this file in the NCS repo:
 - [nRF Connect SDK documentation on built-in Memfault Firmware SDK integration](https://developer.nordicsemi.com/nRF_Connect_SDK/doc/1.8.0/nrf/libraries/others/memfault_ncs.html)
   (starting from v.1.6.0).
 - [Nordic's sample Memfault Firmware SDK integration project](https://developer.nordicsemi.com/nRF_Connect_SDK/doc/1.8.0/nrf/samples/nrf9160/memfault/README.html#memfault-sample).
+
+## Frequently Asked Questions (FAQ)
```
noahp, line 590: i added the text `FAQ` for ctrl+f :shrug: probably redundant tbh

docs/mcu/nrf-connect-sdk-guide.mdx
```diff
@@ -586,3 +586,72 @@ See the list in this file in the NCS repo:
...
+         url: https://github.com/memfault/memfault-firmware-sdk
+         path: modules/lib/memfault-firmware-sdk
+         revision: 0.31.5
+     self:
+       path: my_example_application
+   ```
+
+   Zephyr's `west` tool will use the explicitly listed version of the
+   `memfault-firmware-sdk` module instead of the one specified by the
+   `nrf-connect-sdk` module.
```
chrisc11, line 626: I think I'd just remove this option since 2 achieves the same thing ?

docs/mcu/nrf-connect-sdk-guide.mdx
```diff
@@ -586,3 +586,72 @@ See the list in this file in the NCS repo:
...
+### Updating the Memfault SDK
+
+The nRF Connect SDK embeds the Memfault SDK as a module:
+
+- <https://github.com/nrfconnect/sdk-nrf/blob/v2.0.2/west.yml#L169-L172>
+
+If you would like to update the Memfault SDK without updating the nRF Connect
+SDK here are some strategies:
+
+1. Add the Memfault SDK as a module in the application's `west.yml`:
```
chrisc11, line 601: Maybe: #### so a direct link can be shared?

chrisc11 (approved):
WOO, thanks!

docs/mcu/nrf-connect-sdk-guide.mdx
noahp: ok hear me out... i think option 1 is a _teensy_ bit preferable, because it doesn't rely on the import path being the same in the NCS module list. also it's explicitly listed as an example of how to achieve module forking:
https://docs.zephyrproject.org/latest/develop/west/manifest.html#example-1-3-downstream-of-a-zephyr-release-with-module-fork

vs.

https://docs.zephyrproject.org/latest/develop/west/manifest.html#example-3-3-downstream-with-path-blocklist

but i feel very unstrongly on the topic :sweat_smile:

chrisc11: That's reasonable to me -- then remove option 2? -- Both do ~same thing so just want one option we recommend rather than a few options

PR #279, opened 2022-08-04
MFLT-6704: Trace signature & Issue documentation

tyhoff:


docs/platform/issue-management.mdx
```diff
@@ -14,6 +14,43 @@ the Trace.
 In other words, a Trace is a single occurrence of a problem and an Issue is a
 group of Traces manifesting the same problem.
 
+### De-Duplication of Traces into Issues
+
+A Trace's signature is a SHA256 hash of various pieces of information generated
+by the device as the time the Trace was captured. These pieces of information
```
chrisc11, line 20:
```suggestion
by the device at the time the Trace was captured. These pieces of information
```

docs/platform/issue-management.mdx
```diff
@@ -14,6 +14,43 @@ the Trace.
...
+3  write_something
+2  do something
+1  prv_update_context  <-- New stack frame added in new release
+0  prv_main
+```
+
+If a backtrace for a Trace changed from A to B between two releases, two Issues
+would likely be created. This is because the backtraces changed, and Memfault
+can not be 100% sure they are the same bug. The best practice is to merge the
+Issues that get created but are essentially the same.
```
chrisc11, line 52:
```suggestion
can not be 100% sure they are the same bug. The best practice is to merge the
Issues in this scenario.
```

chrisc11 (approved):

PR #281, opened 2022-08-05
MFLT-6709 Testing an automation again

tyhoff:
Really wanting to see if something works!

asdfasf - new stuff really? WOW! tttt


HBehrens:
Any context on this, @tyhoff ? Should there be some internal signal as a result (e.g. Slack channel)? As you're on it, do you think you can hook up this repo to YouTrack as well?

PR #290, opened 2022-08-17
Add dedicated section about coredump storage test

noahp:
This question comes up enough that I wanted a dedicated section to link
to. Tweak the original location so the content can be shared.


src/pages/_partials/_steps/_coredump-storage-test.mdx
```diff
@@ -0,0 +1,14 @@
+```c
+// Runs a sanity test to confirm coredump port is working as expected
```
chrisc11, line 2: Perhaps add an #include "memfault/components.h" above so someone knows where to pick up functions

chrisc11 (approved):

PR #291, opened 2022-08-23
Clean up duplicate links I added

noahp:
Also update docs link to latest version, 2.0.2, from 1.8.0


fnune (approved):
LGTM. Haven't investigated but we could try to parametrize those URLs so that we store the NCS version we link to in one place...?

noahp:
> LGTM. Haven't investigated but we could try to parametrize those URLs so that we store the NCS version we link to in one place...?
> 
> (OT, not for this PR)

Definitely something we could do!

PR #292, opened 2022-08-23
Update docusaurus to 2.0.1

noahp:
It's no longer in beta, update it:

```bash
❯ yarn upgrade @docusaurus/core@latest @docusaurus/preset-classic@latest @docusaurus/plugin-ideal-image@latest
```

Also update `browserlist`, it was complaining about out of date db:

```bash
❯ npx browserslist@latest --update-db
```


lunelson (approved):
👍🏼

PR #301, opened 2022-09-05
Prepare for Linux SDK Dunfell release: update GH URLs to use - (default branch alias)

martijnthe:
- Updated the "Assumptions and dependencies" section to also mention 3.1 Dunfell.
- Updated all links to the SDK to use `-` instead of `main` as the branch. Github will redirect these links to the default branch (which is set up to be the latest supported Yocto release branch).


docs/linux/introduction.mdx
```diff
@@ -83,9 +83,9 @@ An integration example can be found under
 lives in a Yocto layer in [`/meta-memfault`][source-meta-memfault].
 
 [source-yocto-example]:
-  https://github.com/memfault/memfault-linux-sdk/tree/main/meta-memfault-example
+  https://github.com/memfault/memfault-linux-sdk/blob/-/meta-memfault-example
 [source-meta-memfault]:
-  https://github.com/memfault/memfault-linux-sdk/tree/main/meta-memfault
+  https://github.com/memfault/memfault-linux-sdk/blob/-/meta-memfault
```
fnune, line 88: One detail: `tree` is kind of like a directory on GitHub, and `blob` is a leaf node/file. It's strange, but if I try `tree/-` it doesn't work. But `blob/-` always works, regardless of whether the link is to a blob or a tree. :shrug:

check-linux-sdk-links.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+if grep -r -n --color -e 'memfault/memfault-linux-sdk/\(blob\|tree\)/[^-]\{1\}' docs; then
+  echo "Please use '.../blob/-/...' links! The '-' makes Github automatically redirect to the default branch."
+  exit 1
+fi
```
fnune, line 6: Great idea!

fnune: Maybe forbid `tree/-` since that doesn't work?

fnune (approved):
Awesome!

docs/linux/introduction.mdx
martijnthe: yeah I noticed that too

fnune (approved):

PR #308, opened 2022-09-15
MFLT-7163 Work Around issue with "staged rollout" icon

HBehrens:
As part of #255 we seemingly broke the icon component `<StyledPercentageOutlined />`.
The introduced image optimization leads to different file names as part of the build that are not incorporated in MDX components.

This change works around this issue by moving the referenced PNG to a folder beneath `/static/**` which is ignored by `@docusaurus/plugin-ideal-image`. This change further repeats the steps for the otherwise unaffected component `<NormalizedCharts />` (GIFs are not re-processed) so that all icon components follow the same pattern of having their assets beneath `/static/components/icons`. This aims at consistency so that future authors will hopefully follow this pattern and avoid problems with any image assets.

## Testplan
CleanShot 2022-09-15 at 17 15 04

Observe that icon is now rendered as expected.
[Deep link to respective section (scroll up slightly) on preview deployment](https://deploy-preview-308--memfault-docs.netlify.app/docs/platform/ota/#managing-ota-updates).

CleanShot 2022-09-19 at 12 26 58

Observe the icon for normalized charts still renders as expected for both light and dark mode.
[Deep link to respective section](https://deploy-preview-308--memfault-docs.netlify.app/docs/platform/charts#toggling-chart-normalization-onoff)


HBehrens:
Maybe @lunelson, you have a more robust idea?

lunelson:
> Maybe @lunelson, you have a more robust idea?

@HBehrens sorry for the delay here. You're right that the plugin will optimize all imported raster image files _other than GIFs_, so changing the file type works; however for these kinds of assets I'd rather recommend they be moved to the `static` directory and referred to with a static path. So the icon image file could move like this:

```
.
├── src/
│   └── components
└── static/
    └── icons/
        └── styled-square-percentage.png
```

and the component be changed to this:

```jsx
import React from "react";

export default function StyledPercentageOutlined() {
    return (
        <img
            src="/icons/styled-square-percentage.png"
            style={{ width: "1em", height: "1em" }}
            alt="Blue percentage icon"
        />
    );
}
```

HBehrens:
I like that better, albeit it not being perfect as it tears apart the different aspects of this and future icon components. Still, I will make this change to lead by example so that future authors won't run into this problem again.

HBehrens:
Updated the PR to generalize the pattern you suggested @lunelson (see testplan).

src/components/icons/NormalizedCharts.js
```diff
@@ -1,16 +1,16 @@
 import React from "react";
 
-import iconLight from "./icon-normalized-charts-light-64x64.gif";
-import iconDark from "./icon-normalized-charts-dark-64x64.gif";
+const IconLightUrl = "/components/icons/icon-normalized-charts-light-64x64.gif";
+const IconDarkUrl = "/components/icons/icon-normalized-charts-dark-64x64.gif";
```
HBehrens, line 4: not sure what our coding style says about constants being capitalized and spatially detached from the code that uses it (equally true for `StyledPercentageOutlined`).

mrcljx (approved):

PR #313, opened 2022-09-29
Document the new coredump_plugin.coredump_max_size_kib 

martijnthe:
Adds docs for the upcoming coredumps feature.


docs/linux/reference-memfaultd-configuration.mdx
```diff
@@ -152,6 +154,22 @@ To configure the location where coredumps are stored during processing, see
...
+Next, `memfaultd` will determine the maximum size that is allowed, based on the
+`coredump_max_size_kib`, `storage_max_usage_kib` and `storage_min_headroom_kib`
+configuration values and the amount of available storage space.
+
+Unless there is no available storage, the coredump is written into a temporary
+holding area inside the [`data_dir`][docs-data-dir].
+
+Finally, the coredump is added to the upload queue. This queue is serviced
+periodically (see the top-level `refresh_interval_seconds`).
+
```
fnune, line 172: Sounds a bit more like something that fits in `linux-coredumps` and is linked to from here?

fnune (approved):

docs/linux/reference-memfaultd-configuration.mdx
martijnthe: Hmm yeah on second thought, I'm not sure. That doc is more an integration guide.

I feel like the explanation in the table here is a bit too shallow, so I wanted to elaborate on what the configuration is used for. But then some of these values are used in unison / as part of a larger whole, so explaining them inside the table felt odd.

PR #316, opened 2022-10-06
Add diagram for build ids in symbol files

noahp:


tyhoff (approved):
Is that really an excalidraw image? Do we need to pretty this diagram up?

noahp:
> Is that really an excalidraw image? Do we need to pretty this diagram up?

It is- it's using a non-default font, last one of these I made Heiko elected to go with the nono-default font. I'll switch it back, I think it looks better with the default font!

PR #317, opened 2022-10-12
Further specify how to upload symbols for Yocto builds

fnune:


martijnthe (approved):
Thanks, good call-out.

PR #322, opened 2022-10-21
Update nrf connect sdk guide for Bluetooth

collhar:
Now that we are a part of the Bluetooth SIG, we need to follow their rules of nomenclature. They do not allow "BLE". It needs to be Bluetooth Low Energy or Bluetooth LE.

Also Nordic updated our press release from all Bluetooth to Bluetooth LE, so we should update our docs to match their requests.


noahp (approved):
:heart: wonderful, thank you!

PR #330, opened 2022-11-02
metrics-for-battery-life: fixup charge drop calculations

ejohnso49:
Went through the page again and found a couple of small mistakes in calculations for the charge level drop examples.


chrisc11 (approved):
Thanks @ejohnso49 !

PR #337, opened 2022-11-11
Minification cleanup 

can-memfault:
All the images/gifs with `-min` suffix in the file name and the references from the mdx files are deleted from docs. See the thread which triggered this change: 

image

https://memfault.slack.com/archives/C01APB05QAY/p1668177873715999


fnune (approved):
LGTM. Looks like you also minified some images further? Or you took the `-min` version and renamed it to remove the `-min` suffix?

I'm fine with that.

I checked the live deployment shortly and it looks good. Docusaurus should complain about missing images at build time.

fnune:
@HBehrens thanks for the reminder!

can-memfault:
Ping @fnune :) Do you think what I mentioned in the comment above is alright?

fnune:
> I deleted all the images with the min suffix, and changed all respective references to point to the "non-minified" versions of the images.

I understand :+1: Let me rephrase.

image

As I went through the images I saw that most of the changes were removals. This is expected: you're removing the `-min` images. But two of them were a change in the binary. Those two are paths that do _not_ have the `-min` suffix. So my question was: what happened to those? Did you minimize them further?

I was just curious, I think the PR should be merged either way :)

PS Sorry that I missed the notification again :disappear:

can-memfault:
Ah now I got the question @fnune :) I "Figma'd" them because they were missing the new "Weekly" and normalization buttons in the header.

PR #365, opened 2022-12-23
Add section on "Allowing Downgrades" to OTA documentation

martijnthe:


docs/platform/ota.mdx
```diff
@@ -221,6 +221,25 @@ An example:
   />
 </p>
 
+### Allowing Downgrades
+
+For Delta Releases, downgrades can be accomplished by creating a new Delta
```
HBehrens, line 226: Did we validate that this works even if the Version Check as described below is still on?

docs/platform/ota.mdx
```diff
@@ -221,6 +221,25 @@ An example:
...
+Release that has the newer version as its "From" version and the older version
+as its "To" version.
+
+For Full Releases, downgrades are disallowed by default. This is to prevent a
+device from downgrading to a version that is known to be buggy. It is possible
```
HBehrens, line 231:
```suggestion
device from downgrading to a version that is known to be erroneous. It is possible
```

HBehrens (approved):
I am not 100% sure the behavior we describe for Delta Releases what's currently implemented (see inline comment). Plus one minor language remark.

Otherwise, LGTM

docs/platform/ota.mdx
martijnthe: I have not validated it. I added this fully based on the assumption that this statement from you is correct:

> Stating the obvious: When using delta releases for downgrades, you will have to have one “route” per high version you want to downgrade from.

But reading [this ticket](https://memfault.myjetbrains.com/youtrack/issue/MFLT-7887) makes me wonder whether it's supported at all! 

@mrcljx any wisdom to share?

mrcljx: https://memfault.slack.com/archives/C03EFSM0LBU/p1667991354062629?thread_ts=1667987155.220839&channel=C03EFSM0LBU&message_ts=1667991354.062629

It's indeed not possible at the moment. There is a constant currently that prevents this.

I personally have would love to always allow downgrades through delta releases: if someone creates a 2.0 to 1.0 release, I find this quite explicit.

So far @HBehrens has rejected this, because it's possible to create loops that way. And validating that the update graph is loop free is a bit more complicated, and we never had the time to implement that check.

martijnthe: Thanks @mrcljx. Updated to say "For Delta Releases, downgrades are currently not supported."

PR #370, opened 2023-01-06
Fixup scroll margin on anchors

ejohnso49:
## Summary

When using anchors that point to elements that are not headers, the navbar hides these. This change updates the selector to targets to filter on all target elements (not just headers) and uses scroll-margin-top instead.

Example:

anchor_navbar_hidden

W/ Fix:
footnote_anchor_margin


fnune (approved):
TIL `:target` and `scroll-margin-top` :smile: Looks like `:target` is an ancient feature but `scroll-margin-top` is fairly modern. That's OK :+1: 

Maybe go over old "call"-sites of `.anchor-offset` and remove them? Should be only `memfault-terminology.mdx`.

docs/platform/memfault-terminology.mdx
```diff
@@ -5,41 +5,41 @@ sidebar_label: Memfault Terminology
...
+| Symbols                                        | The .elf file that is used by Memfault to symbolicate any crash data.                                                                                                                                                                                                                                                                                                                                                                  |
+| Trace <a id="trace"></a>                       | A crash or snapshot of the device at a particular moment in time, which can later be converted to a backtrace by Memfault. A Trace can be one of three types: - Trace Event - captures only the program counter (instruction pointer) and the link register (return address) if available - Stack Trace - captures the memory regions of the stack - Coredump Trace - captures most, if not all, relevant memory regions of the system |
```
noahp, line 45: dangit, i thought i had added an anchor tag so these are visible without inspecting :disappointed: oh well rip

ejohnso49: I have one remaining issue still even with this change. I cannot for the life of me get the page to scroll to the right spot when navigating from a link on another page, it's like it's treating the `<a>` as `inline` no matter what I do, even with a class rule to set it explicitly. If I refresh/go straight to URL with the page fragment it scrolls correctly.

PR #380, opened 2023-01-17
Add a note about trimming captured TCB size

noahp:
Also drive by update the heading levels on the RTOS analysis page.


ejohnso49 (approved):
Love it! Thank you!

src/pages/_partials/_steps/_rtos-ports.mdx
```diff
@@ -49,4 +49,65 @@ int memfault_platform_boot(void) {
...
+
+If the platform is configured to not capture all of RAM (due to storage or
+bandwidth limitations), it's still possible to capture the active FreeRTOS tasks
+as part of a coredump. See
+[this file](https://github.com/memfault/memfault-firmware-sdk/blob/0.37.1/ports/freertos/src/memfault_freertos_ram_regions.c)
+for details on that implementation.
+
+With that in place, the Memfault SDK will selectively capture the running tasks
+during a coredump- both the Task Control Block (TCB) structure as well as a
+limited backtrace for each task.
```
tyhoff, line 62:
```suggestion
With that in place, the Memfault SDK will selectively capture the running tasks
during a coredump by saving both the Task Control Block (TCB) structure as well as a
limited backtrace for each task.
```

src/pages/_partials/_steps/_rtos-ports.mdx
```diff
@@ -49,4 +49,65 @@ int memfault_platform_boot(void) {
...
+below. Memfault requires the struct members through `pcTaskName` to decode the
+task.
+
+```bash
+# note- the 'tskTaskControlBlock' is the actual type name for the 'TCB_t' struct
+❯ pahole -C 'tskTaskControlBlock' ./build/memfault-esp32-demo-app.elf.memfault_log_fmt
+```
+
+The unused `xNewLib_reent` member below is 240 bytes out of the total 352 byte
+struct size.
```
tyhoff, line 80: You can give a firmer example why the user might want to trim it down. 

For example, above - 

Under some configurations, the TCB structures consume significant amounts of
coredump space (300-1000 bytes per TCB), such as when newlib is configured to have a reent struct for each task. This can cause coredumps to become much larger than necesary.

To mitigate this problem...

...

In the printout below, the the `xNewLib_reent` member is 240 bytes out of the total 352 byte
struct size.

src/pages/_partials/_steps/_rtos-ports.mdx
```diff
@@ -49,4 +49,65 @@ int memfault_platform_boot(void) {
   memfault_build_info_dump();
 ```
 
+#### Configure FreeRTOS Task Capture Size
```
tyhoff, line 52: It sounds to me like we *always* want to configure the TCB capture size to 100 bytes. Any reason why not make this the default? 

Memfault should provide sane defaults and let power users opt-out, rather provide a semi-working default and allow power users to fix brokenness.

tyhoff (approved):

src/pages/_partials/_steps/_rtos-ports.mdx
noahp: yeah... definitely agree, i'll make that change!

ejohnso49 (approved):

tyhoff:
Would love to see all of this be default ON and sane (100 bytes, prioritize tasks by default, ignore the extra fields, etc) and then the note just mentions how smart our SDK is.

The doc is good to go if we need something out.

noahp:
Ok, `0.39.0` of the SDK updates the default TCB size, and I've updated this PR to match.

ejohnso49 (approved):

PR #384, opened 2023-01-23
MFLT-8803 Document deleting a release

tyhoff:
Customers are getting tripped up when deleting Releases because they have outstanding Activations that need to be marked as Pulled.

Document this fact in our docs page.


noahp (approved):
:heart_eyes:

victorbordo (approved):

PR #388, opened 2023-01-27
[WIP] doc on tracking stack and heap usage

noahp:


docs/mcu/memory-tracking.mdx
```diff
@@ -0,0 +1,151 @@
...
+id: memory-tracking
+title: Tracking Stack and Heap Usage
+sidebar_label: Tracking Stack and Heap Usage
+---
+
+import IntegrationCompletePrerequisite from "@site/src/pages/_partials/_integration-complete-prerequisite.mdx";
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+<IntegrationCompletePrerequisite />
```
tyhoff, line 11: This being placed in so many MCU pages feels a bit aggressive. Is this really helping people? 

Fine to leave it in though - we do it everywhere else.

docs/mcu/memory-tracking.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+### RTOS Heap Usage Metric
+
+RTOS's usually have APIs for measuring heap usage, for example:
+
+**FreeRTOS:**
+
+- [xPortGetFreeHeapSize](https://www.freertos.org/a00111.html#heap_4)
+
+**Zephyr RTOS:**
```
tyhoff, line 114: Not sure how you want to weave it in, but we collect the heap bytes free already in a bunch of metrics. For example, Zephyr: https://github.com/memfault/memfault/blob/5df3a95de0c0474df6306da6663300434704e2cd/sdk/embedded/ports/zephyr/common/memfault_platform_metrics.c#L78

Maybe use the tab view to show which systems have it enabled and show the bare metal usage below for systems that don't already have the stat built in.

tyhoff (commented):
Related, but not required for your changes. It would be nice to mention / instruct how to capture Zephyr heap stats on the other heap stats page. 

https://github.com/memfault/memfault/pull/7852/files#diff-652a4fa208dc2495076bfc2107d94bc40637d9ae8580715465d5e39b46b26260R81

Is there a reason to have this page be separate from the existing heap one? I tend to think we should have a page on tracking stack, and a separate on tracking heaps, and combine both heap tips into that page.

docs/mcu/memory-tracking.mdx
noahp: hah yeah, definitely agree. maybe we should reduce that to a little note version instead of the big blob, i'll put something up

PR #390, opened 2023-02-06
Expand a bit on trace event variants

noahp:


ejohnso49 (approved):
Might be good to mention that different status values will generate unique issues while different logs will not.

chrisc11 (approved):

ejohnso49 (approved):

PR #393, opened 2023-02-13
Add info text for metric display name

can-memfault:
@HBehrens it'd be great to review this soon so I can include it in the January release notes


HBehrens (approved):
made a small suggestion, otherwise LGTM

PR #395, opened 2023-02-13
Handle Auto Dark Mode

noahp:
Chrome's "Auto Dark Mode" will run all over your nice theme, even if you
have a nice dark mode enabled 😩

Theoretically we can disable it with this meta tag but my Chrome (Google
Chrome 110.0.5481.77) did not respect it:

https://developer.chrome.com/blog/auto-dark-theme/#using-a-meta-tag

Instead, globally disable with the css `:root` setting.

I also deleted some stuff @fnune that seemed to have no effect on theme
toggle, LMK if that was wrong!

Results:

- before, light mode / Auto Dark Mode (enabled via dev tools):

   

- after:


src/components/EmbedMode.js
```diff
@@ -18,46 +18,10 @@ export function EmbedModeProvider({ children }) {
                         <meta name="robots" content="noindex" />
                     </Head>
                     <div className="embed-mode">{children}</div>
-                    <ForceColorMode mode="light" />
```
fnune, line 21: Note that this is in `EmbedMode`: the docs are wrapped in `EmbedMode` when they show in the Memfault app as an iframe. This element is here to prevent showing dark mode docs in a light-mode (only) web app.

I really wanna get rid of that iframe, but for now, let's keep this in.

Maybe add a comment? It's not very clear from the code I wrote a while ago.

fnune (changes requested):
Requesting changes because `ForceColorMode` needs to remain (unfortunately).

Also, the fix isn't working for me...?

Here's "light" mode:

image
image

I tried hacking around several things like you probably did (all the [options in the Chrome docs][0]), but it didn't help. My Chrome version (Chromium):

```
Version 110.0.5481.77 (Official Build) (64-bit)
```

[0]: https://developer.chrome.com/blog/auto-dark-theme/#using-a-meta-tag

src/components/EmbedMode.js
noahp: 🤦 oh duh, makes sense! Thank you, will do

noahp:
> Also, the fix isn't working for me...?
> 
> Note that I had to enable auto-dark mode via Chrome flag, so maybe in my build respecting the opt-out isn't implemented yet.

dangit, you're right :( when I was using the dev tool toggle, it also set `prefers-color-scheme`, which overrode the Auto Dark Mode changes.

It looks like the "problem" happens when Auto Dark Mode is enabled AND `prefers-color-scheme` is unset or `light`. It's really annoying that those options in the Chrome docs don't appear to work :(

PR #396, opened 2023-02-16
MFLT-9126 Remove CDR Early Access message

victorbordo:
### Summary
Remove the EA message now that CDR is globally available.

### Test plan
Check message is removed in PR build.


noahp (approved):
SGTM, but let's confirm @chrisc11 is cool with the language change :+1:

PR #403, opened 2023-02-23
Add note about many:1 sw_version:symbol_file

noahp:
In the NRF5 SDK DFU, you can change the device's fw version without
touching the application symbol file- the fw version is written to the
"bootloader settings page" in flash, and retrieved at runtime, rather
than baked into the application.

Add a note about this case and how it's modeled in Memfault.


docs/mcu/symbol-file-build-ids.mdx
```diff
@@ -137,3 +138,19 @@ memfault --org-token $ORG_TOKEN --org acme-inc --project smart-sink \
   --revision 89335ffade90ff7697e2ce5238bd4c68978b6d6e \
   build/symbols.elf
 ```
+
+## Software Version vs. Build ID in Memfault
```
noahp, line 142: i'm not thrilled about the wording in this section, definitely open to re-writing it (or adding a diagram)

docs/mcu/symbol-file-build-ids.mdx
```diff
@@ -137,3 +138,19 @@ memfault --org-token $ORG_TOKEN --org acme-inc --project smart-sink \
...
+
+On the Memfault backend, Software Version and Build ID have different purposes:
+
+- **Software Version** is attached to Device data (both to populate the Device
+  entity itself, as well as tagging individual reports from a Device), and can
+  be used when filtering various views/data in the Memfault web UI.
```
ejohnso49, line 148: Might be worth highlighting that software version is used with OTA/cohorts too

ejohnso49 (approved):

chrisc11:
I might reframe this as a "Frequently Asked Questions" -- here's a gist that needs some editing

1. When using a Build ID, is it okay if there is not always a 1:1 mapping with Software Version?

Yes, in fact this is one of the benefits of using BuildIds. In this scenario there will always be a 1:1 mapping between the binary running and a Build Id so Memfault is guaranteed to recover accurate stack traces while at the same time giving you flexibility around how you report software versions

Two common situations where there is not a 1:1 mapping are the following:

  a) "Debug" builds -- Memfault recommends using semver and for pre-release debug builds using a common software version (i.e 1.1.1-dev). This way all the crashes and issues observed during debug get bucketed under a single version that's easy to track in the UI. In this situation you wind up with many separate BuildId mapping to a single SoftwareVersion
  b) [nrf5 example here] In this situation a single build id will map to multiple software versions.

docs/mcu/symbol-file-build-ids.mdx
```diff
@@ -137,3 +138,19 @@ memfault --org-token $ORG_TOKEN --org acme-inc --project smart-sink \
...
+- **Build ID** is used only for decoding Device data (coredumps, traces, etc)
```
tyhoff, line 149: I would almost specify build id's first, and then define software version as the 'everything else'. 

- **Build ID** is only used for decoding Device data (coredumps, traces, etc).
- **Software Version** is used for everything else. It is attached to Device data, used for OTA, and used to filter Devices by last seen Software Versions, etc.

noahp:
> I might reframe this as a "Frequently Asked Questions" -- here's a gist that needs some editing
> ...

Sweet, I used this- I think it's a lot better!

PR #420, opened 2023-03-07
MFLT-9320 Document use of LTO vs coredumps

martijnthe:
Added a sentence to the "prerequisites" of the Linux coredumps guide.


docs/linux/coredumps.mdx
```diff
@@ -153,6 +153,12 @@ USER_CLASSES = "buildstats"
 [prelinking-example]:
   https://github.com/memfault/memfault-linux-sdk/blob/-/meta-memfault-example/conf/local.conf.sample#L5
 
+### Link Time Optimization (LTO)
+
+We strongly recommend _against_ using Link Time Optimization (LTO) in the
+programs you want to be able to debug using coredumps. When LTO is enabled, the
+analysis of the coredumps can become lacking or even fail completely.
+
```
martijnthe, line 161: The MCU side of the docs may also benefit from a blurb like this, but I wasn't sure where to put it.
cc @noahp

sarfata (approved):
I know we had problem with LTO in the context of rust but how does this manifest when it's not a rust binary?

martijnthe:
> I know we had problem with LTO in the context of rust but how does this manifest when it's not a rust binary?

The advice likely still applies because LTO can shuffle code so much around that a coredump/backtrace becomes hard to map bcak to the original source code.

PR #443, opened 2023-04-06
Move the MCU FAQ section to its own page

noahp:
And add a note about `-fvar-tracking` error when building with debug
info disabled.


docs/mcu/faq.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+The following error may occur when building an application that includes the
+Memfault Firmware SDK:
+
+```bash
+cc1: error: variable tracking requested, but useless unless producing debug info [-Werror]
+```
+
+The Memfault SDK requires building with **debug info** enabled. Note that
+enabling debug info does not impact the binary executable in the final program,
```
chrisc11, line 92: good callout

chrisc11 (approved):

PR #448, opened 2023-04-12
Mention memfaultd should start before fluent-bit

martijnthe:


martijnthe:
Found that fluent-bit only retries once by default:
https://docs.fluentbit.io/manual/administration/scheduling-and-retries#configuring-retries

Increasing that a bit is probably the recommendation we should make.

PR #450, opened 2023-04-13
MFLT-9323 Update public docs with the explanation of new Range implementation

martijnthe:
localhost_3000_docs_platform_metrics_


docs/best-practices/metrics-for-battery-life.mdx
```diff
@@ -214,7 +217,7 @@ static void prv_metrics_heartbeat_record_battery_info(void) {
         );
     } else {
         // Don't report this metric during hours with charging events
-        // Make sure to set up -1 as the value to ignore in the Memfault service
+        // Make sure to set up the range & ingress filtering to ignore -1 in the Memfault service
```
martijnthe, line 220: @tyhoff @ejohnso49 jfyi, I noticed this example does not make use of [the new "ignore NULL" MCU heartbeat feature](https://github.com/memfault/memfault-firmware-sdk/blob/master/CHANGES.md#changes-between-memfault-sdk-0420-and-0412---mar-22-2023).

ejohnso49: I can submit a separate PR to update this. Thank you!

docs/platform/metrics.mdx
```diff
@@ -288,6 +290,51 @@ To filter by Timeseries, add a **Historical Data** filter and select
...
+| :-------------------------------------------------------------- | ----------------- | ------------------------------- |
+| [Fleet-wide Metrics charts][metric-charts]                      | 🟢 Included       | ⛔ Ignored                      |
+| [Device Search][device-search]                                  | 🟢 Included       | ⛔ Ignored                      |
+| [Device-Threshold Alerts][device-alerts]                        | 🟢 Included       | ⛔ Ignored                      |
+| [Fleet-Threshold Alerts][fleet-alerts]                          | 🟢 Included       | ⛔ Ignored                      |
+| Device Timeline: Heartbeat data                                 | 🟢 Included       | ⛔ Ignored                      |
+| [Device Timeline: High Resolution Telemetry (HRT)][android-hrt] | 🟢 Included       | 🟢 Included (known issue)       |
+| MCU Device Metrics Charts                                       | 🟢 Included       | 🟢 Included (known issue)       |
+| Raw Data Exports                                                | 🟢 Included       | 🟢 Included                     |
```
vlindhol, line 328: Something doesn't fit right with me with how we are presenting this feature (not only in these docs). The setting is called "Ingress filtering -> Accept only values within range" and a hint saying explicitly that data is not being ingested at all. We of course know that we're not dropping anything, but we probably need to be consistent with how this is communicated the customer.

We could:

1. Keep pretending we are dropping data. We can then mention that due to some technical reasons, the dropped data will still be visible in features X, Y and Z.
2. Change the wording around this setting in the product so customers know we keep the data but are not showing it. Maybe something as simple as "Values outside of range should be ignored: Yes/No" with some helper text explaining that the data is still ingested.

Anyhow, this is more of a product consideration, I'll write a follow-up comment with some thoughts about this particular PR :)

vlindhol: A suggestion in line with suggestion (1) above: a table is always a bit more of a cognitive load to start scanning through. Would it be more friendly to just list the "exceptions to the rule"? Something like replacing the table with:

```suggestion
- "Accept all values": all values are accepted, regardless of whether they are
  within the range or not. This is the default for new metrics. The range may be
  used for display purposes in the future.
- "Accept only values within range": only values within the range are accepted. Values outside the range are dropped from most Memfault features.

::: info

**Raw data exports** still include all values, regardless of the setting above. For technical reasons, all data is currently also shown in **MCU Device Metrics Charts** and [High Resolution Telemetry in the Device Timeline][android-hrt].

:::
```

These diffs are hard to read: I changed the wording in the second list item and replaced the table entirely.

vlindhol (approved):

docs/platform/metrics.mdx
martijnthe: @HBehrens via slack:

> I believe what we are communicating with the table is correct: Our goal (only partially implemented, see table) is to exclude certain values from everything but the raw data.
I agree that we could improvide on how we communicate that. Maybe instead of talking about “accepting” vs. “dropping”/“ignoring” we could focus on the “processing” part? After all, we do accept and store the data. All we’re doing here is preventing unexpected values (those that are outside the anticipated range) to distort the data as we process it.
At this point, I’d focus on improvements that are easy to make though ([e.g. no](http://e.no/) major re-wording in the UI and by extension docs) to get this feature out. What we’re doing is no different from what we always did.
You proposed such a change by removing the table. I’d personally keep it as we saw customers (and CSM) asking exactly at that level of granularity (“how about this screen?“). If we keep the table, I’d recommend being more explicit about the unimplemented areas being a bug/soon to be removed.

docs/platform/metrics.mdx
```diff
@@ -288,6 +290,51 @@ To filter by Timeseries, add a **Historical Data** filter and select
...
+#### Effect of "Ingress Filtering" on out-of-range values, per feature
+
+| Feature                                                         | Accept all values | Accept only values within range |
+| :-------------------------------------------------------------- | ----------------- | ------------------------------- |
+| [Fleet-wide Metrics charts][metric-charts]                      | 🟢 Included       | ⛔ Ignored                      |
+| [Device Search][device-search]                                  | 🟢 Included       | ⛔ Ignored                      |
+| [Device-Threshold Alerts][device-alerts]                        | 🟢 Included       | ⛔ Ignored                      |
+| [Fleet-Threshold Alerts][fleet-alerts]                          | 🟢 Included       | ⛔ Ignored                      |
+| Device Timeline: Heartbeat data                                 | 🟢 Included       | ⛔ Ignored                      |
+| [Device Timeline: High Resolution Telemetry (HRT)][android-hrt] | 🟢 Included       | 🟢 Included (known issue)       |
```
chrisc11, line 326: would remove "known issue"

chrisc11 (approved):

PR #452, opened 2023-04-13
fluent-bit: recommend Retry_Limit override

martijnthe:


martijnthe:
@sarfata ping

docs/linux/logging.mdx
```diff
@@ -106,6 +106,18 @@ This is the required output configuration:
...
+limit is 1. We recommend setting to a higher value to decrease the chance of
+losing logs in the event that memfaultd is (re)starting while fluent-bit is
+attempting to flush logs:
+
+```
+[OUTPUT]
+	Name tcp
+	...
+	Retry_Limit 5
+```
```
sarfata, line 119: I would just include what you have added to fluent.conf into the code snippet above: If we think it should always be in for things to work (we do) then it should be part of our "required" config.

This is the `[OUTPUT]` block with comment:
```
[OUTPUT]
	Name tcp
	Host 127.0.0.1
	Port 5170
	Format msgpack
	Match *
	net.keepalive on
	net.keepalive_idle_timeout 10
	# Default retry limit is 1. We recommend setting to a higher value to
	# decrease the chance of losing logs in the event that memfaultd is
	# (re)starting while fluent-bit is attempting to flush logs:
	Retry_Limit 5
```

Sorry I cannot offer a "suggestion" that does this for you (too far up + github bugs with the code blocks inside suggestion block).

sarfata (approved):

PR #454, opened 2023-04-14
MFLT-9625 Update metrics-for-battery-life with changes from v0.42.0

ejohnso49:
From #450 I have create a new PR to update the battery metrics page with updates related to v0.42.0. I also found a few issues in our sample code. I am not completely satisfied with my methods for splitting out changes pre and post v0.42.0. Any feedback there would be highly appreciated.


docs/best-practices/metrics-for-battery-life.mdx
```diff
@@ -108,7 +122,9 @@ This value should be based on a linear scale. Millivolts should not be sent.
 - To report a battery charge level of 50 mAh out of a 500 mAh capacity battery,
   record 1,000.
 - If the battery charge level could not be acquired or it is known to be
-  invalid, record -1.
+  invalid, record -1 _(SDK < v0.42.0)_.
```
bahildebrand, line 125: What would happen if someone used `-1` on an SDK version >= `0.42.0`? If the outcome is particularly gnarly should we call this out in the changelog loudly? (not sure if we have already)

bahildebrand (approved):

docs/best-practices/metrics-for-battery-life.mdx
ejohnso49: It won't act any differently than the prior versions, data will be sent and then filtered by the backend according to the range defined. I think the motivation is to highlight the best practice for the two cases. I'm not entirely sure if my changes are the best way to do that though

docs/best-practices/metrics-for-battery-life.mdx
```diff
@@ -178,6 +202,64 @@ metric within the UI.
...
+
+    if (report_battery_drop) {
+        memfault_metrics_heartbeat_set_signed(
+            MEMFAULT_METRICS_KEY_DEFINE(Battery_ChargeLevelDrop),
+            battery_charge_drop
+        );
+    }
+
+    // Also record the current battery charge level
+    if (battery_charge_level < 0) {
```
tyhoff, line 250: Do you mean `battery_charge_level > 0`

tyhoff (approved):

docs/best-practices/metrics-for-battery-life.mdx
ejohnso49: Yes, definitely. Thank you!

PR #459, opened 2023-04-25
Add auxiliary symbol file faq section

noahp:
From original doc:

https://memfault.notion.site/Merging-auxiliary-symbol-files-into-an-ELF-5bbbe5597d2e4903887454e8f3093738

I added an example when there's just 1 ROM elf to include.


martijnthe (approved):

chrisc11:
ty!

PR #463, opened 2023-05-09
Add rough Memfault coredump structure

tyhoff:
I've had a number of times customers asking me why their coredumps are not being parsed correctly only to find out that they are sending us an invalid coredump. A tool is the ultimate goal to distribute to customers, but for now, we can point to the format documentation and have them use hexdump to potentially self-diagnose.

The goal here is not to be exhaustive, but at least document the header, footer, and block structure enough so that people can easily see when something is wrong.


docs/mcu/coredumps.mdx
```diff
@@ -455,3 +455,39 @@ To run the test, call these functions either from your device's CLI, via a
...
+|            | Length        | 4            | Length of block                                                                                                                 | 1280                    |
+| Block 2..N | ...           | ...          | ...                                                                                                                             | ...                     |
+| Footer     | Magic         | 4            | Hard-coded magic bytes in footer                                                                                                | `0x504D5544` (DUMP)     |
+|            | Flags         | 4            | Miscellaneous flags of the coredump                                                                                             | All zeroes              |
+|            | Reserved      | 8            | Reserved for future use                                                                                                         | All zeroes              |
+
+Block's can have the following types:
```
noahp, line 479:
```suggestion
Blocks can have the following types:
```

noahp (approved):
Oh this is sweet! Love it

PR #466, opened 2023-05-09
Distinguish Issue and Metric notifications

noahp:
Add some details on Issue notifications, and rename the sidebar for
Alerts to indicate they're specific to Metrics


docs/platform/issue-management.mdx
```diff
@@ -134,3 +134,18 @@ are displayed.
...
+- Creation
+- Comment added
+- Modified
+- Re-open
+
+These emails can be disabled from the
+[User Settings](https://app.memfault.com/organizations/-/projects/-/settings)
+panel:
+
+![User Notification Settings](/img/docs/platform/user-notification-settings.png)
```
ejohnso49, line 151: Nit: Is possible to center this? If it's not easy, that's okay. I think some future docs work around a standard tag to include images with the markup/properties we standardize would be good

ejohnso49 (approved):
I think this can be merged but we should work to improve the consistency of the terms we use around notifications/alerts. I don't think that should be included as part of this though

docs/platform/issue-management.mdx
noahp: Oh yeah! and 100% agree, we could try to expand ImageFigure a bit, I think it needs a few modifications (ability to set max width explicitly in the thumbnail)

PR #467, opened 2023-05-10
Update notifications->default doc section

noahp:


fnune (approved):
Nit: assuming that big picture renders very wide and long in the page because it's in portrait format.

May be worth giving it a max width or height to keep it reasonable?

I haven't seen the page live, so just complaining blindly really. Approved!

ejohnso49 (approved):
Same comment as #466 , good to approve but we should improve our language consistency between notifications/alerts/etc

PR #469, opened 2023-05-12
Add block data entry to coredump structure table

noahp:


tyhoff (approved):
Ahh - yeah, that's an oversight on my end. Thanks!

PR #476, opened 2023-05-17
Update Kconfig flags for zephyr guide

noahp:
There's a few that are no longer needed for a vanilla Kconfig
integration, clean them up.


docs/mcu/zephyr-guide.mdx
```diff
@@ -57,13 +57,11 @@ supported development repositories here:
...
 ## Enabling Memfault
 
-To enable Memfault, there are a few required Kconfig options that should be set:
+To enable Memfault, set the required Kconfig options:
 
 ```bash
 # Memfault configuration
 CONFIG_MEMFAULT=y
-CONFIG_MEMFAULT_HTTP_ENABLE=n
-CONFIG_MEMFAULT_ROOT_CERT_STORAGE_CUSTOM=y
```
ejohnso49, line 66: Any idea what the reason for these was in the past?

ejohnso49 (approved):

docs/mcu/zephyr-guide.mdx
noahp: Yep- our zephyr port initially only really supported the nrf9160, so for other targets these needed to be n-selected. We made them conditional a while back but I forgot to update this guide

PR #493, opened 2023-06-06
Remove unused optimize-images script

nihalgonsalves:


package.json
```diff
@@ -17,10 +16,8 @@
         "@docusaurus/core": "^2.3.1",
         "@docusaurus/plugin-ideal-image": "^2.3.1",
         "@docusaurus/preset-classic": "^2.3.1",
-        "@endiliey/react-ideal-image": "^0.0.11",
```
fnune, line 20: Is this one related? It seems like it's unused but I didn't know about it.

fnune (approved):

package.json
nihalgonsalves: Yeah, strictly should be in a separate PR but thought I'd also remove two unused deps (see second commit)

fnune: That's alright, just wanted to make sure we didn't break stuff. Looks like we use the other `ideal-image` plugin in the deps.

PR #515, opened 2023-06-29
Add renames and redirects for esp32, esp8266, NCS

noahp:
Follow up from #506, #507, #508, rename everything and add redirects for
external backlinks.


noahp:
Redirects are working, for example:
https://deploy-preview-515--memfault-docs.netlify.app/docs/mcu/esp32-guide/

tyhoff (approved):
Love it.

PR #524, opened 2023-07-10
Fix Linux sdk link returns a 404

rksawyer:
It appears the github urls no longers support a suffix of blob/-/ for the project root.

### DIFF
--- a/docs/linux/quickstart-raspberrypi.md
+++ b/docs/linux/quickstart-raspberrypi.md
@@ -47,7 +47,7 @@ The instructions on this guide are compatible with Linux and macOS (Intel and
 Apple Silicon). Windows users should be able to follow along with minimal
 adjustments.
 
-[sdk]: https://github.com/memfault/memfault-linux-sdk/blob/-/
+[sdk]: https://github.com/memfault/memfault-linux-sdk/
 
 ### Create a Docker container to build with Yocto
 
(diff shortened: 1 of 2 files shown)


sarfata:
> It appears the github urls no longers support a suffix of blob/-/ for the project root.

That is unfortunate - we have a lot more URLs that use the prefix. We even have a linting rule to make sure we use them. I will fix them as part of the 1.5 release too.

sarfata:
Oh I misunderstood you - sorry. It's only the root that does not work anymore. Other URLs still work. You correctly fixed all of them. Thanks a million!

sarfata (approved):

rksawyer:
@sarfata The lint job failed.  Maybe a permissions issue because this was my fork?

rksawyer:
@sarfata I can try to give you permissions to my fork.  Expectedly, I do not have permissions to run CI jobs in the memfault account.  But maybe the job will inherit your permissions for the fork.

rksawyer:
Another option, you can make a branch within the main git repo using the patch.
https://patch-diff.githubusercontent.com/raw/memfault/memfault-docs/pull/524.patch

sarfata:
Thanks @rksawyer I will figure this out tomorrow.

sarfata:
@rksawyer sorry I did not figure out the error. It's inside the GitHub Checkout action which is very weird. Should work for PR from other repositories. I have gone ahead and merged the fix in #526.

PR #528, opened 2023-07-13
Update releases-integration-guide.mdx

Natasha-Co:
To find out where the latest release endpoint is please use api/v0/releases/latest/url. as described here. Please note that your device serial number will be the main identifier that will be used as the main query parameter. To check for the latest release on a device, you can look at the READMEs in your our the [iOS] (https://github.com/memfault/memfault-cloud-ios) or [Android] (https://github.com/memfault/memfault-cloud-android) repos for examples of how a phone app can do this.

### DIFF
--- a/docs/mcu/releases-integration-guide.mdx
+++ b/docs/mcu/releases-integration-guide.mdx
@@ -181,6 +181,8 @@ see. Let's assume we have the following test device:
 - `current_version`: 0.0.1
 - `device_serial`: DEMOSERIAL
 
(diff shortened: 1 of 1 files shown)


docs/mcu/releases-integration-guide.mdx
```diff
@@ -181,6 +181,14 @@ see. Let's assume we have the following test device:
 - `current_version`: 0.0.1
 - `device_serial`: DEMOSERIAL
 
+To find out where the latest release endpoint is please use
```
ejohnso49, line 184: I think we should put this into an FAQ section toward the bottom of the page. Something like this:

```suggestion
## FAQ

### How can I integrate my BLE device with the OTA subsystem?

To integrate you will first need the following:

* An update mechanism to perform the upgrade on your device (eg. MCUboot/SMP)
* A BLE gateway (phone, hub, etc) to query the latest release endpoint on behalf of the device

To query for info on the latest release, use the endpoint as described in the [previous section](#query-for-ota-payload). 
```

PR #537, opened 2023-07-27
Update some code snippets in the bat. life doc

noahp:


tyhoff (approved):
Gahhhh we really need some helpers and a real example app for this metric. 

Thanks for patching this up.

noahp:
> Gahhhh we really need some helpers and a real example app for this metric.
> 
> Thanks for patching this up.

Could not agree more 🤗

PR #538, opened 2023-07-27
fix: battery charge example code errors

bahildebrand:
The example code in the battery charge tracking best practices had some
errors. Fix these so that the code is more accurate to what an end setup
would look like.


bahildebrand:
* main
  * **PR #538** <a href="https://app.graphite.dev/github/pr/memfault/memfault-docs/538" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/memfault/memfault-docs/538?utm_source=stack-comment).

PR #549, opened 2023-08-18
fix: update dead logging code links

bahildebrand:
Two links in the logging subsystem guide were either incorrect or dead.
Replace them with perma-links so that they don't move again as the SDK
is updated.


ejohnso49 (approved):
Definitely an improvement. It would be amazing to somehow flag that these need to be changed if the file on github changes too...

PR #550, opened 2023-08-29
Add why it is recommended to source the Yocto build

Natasha-Co:
Added: 

Why it is recommended to source the Yocto build environment script using source oe-init-build-env before running upload-yocto-symbols.


sarfata (approved):
🍾

PR #553, opened 2023-08-30
Note about capturing freertos variables

noahp:


docs/mcu/rtos-analysis.mdx
```diff
@@ -57,6 +60,122 @@ To enable stack usage high water marking, the `configCHECK_FOR_STACK_OVERFLOW`
...
+define block memfault_freertos { section .bss* object *tasks*.o*, section .bss* object *timers*.o* };
+```
+
+And then update `memfault_platform_coredump_get_regions()` to include that
+block:
+
+```c
+// In the source file where memfault_platform_coredump_get_regions is defined add the following snippet
+// Be sure to include an additional sMfltCoredumpRegion in your regions array for the FreeRTOS region
+// created by the linker script
```
ejohnso49, line 153:
```suggestion
// Be sure to include additional sMfltCoredumpRegion entries in your regions array for the FreeRTOS regions
// created by the linker script and helper function
```

ejohnso49 (approved):
Thank you! Amazing

PR #554, opened 2023-09-04
fix(linux): nasty typo in yocto syntax

sarfata:
The proper syntax to configure a package from `local.conf` is `PACKAGECONFIG:pn-...` (not an underscore).


martijnthe (approved):
Oof..

PR #564, opened 2023-09-13
fix(ImageFigure): Align the caption text with image center

ejohnso49:
The caption text currently is left-aligned. Making these center-aligned makes it easier to understand that they are captions belonging to the image.

Before:
image

After:
image


noahp (approved):
:heart: thank you! great spot

PR #565, opened 2023-09-13
Add warning about using SWT for build variants

martijnthe:
- We had been mentioning that SWTs can be used for build variants, but this is actually not recommended because it does not work well with the "primary SWT" concept.
- Add a note about OTA Release versioning and its relation to the primary SWT.
image


noahp (approved):

docs/platform/software-version-hardware-version.mdx
```diff
@@ -55,8 +55,19 @@ implement as part of the integration.
...
-distinguish between build variants (i.e. `main-mcu-dvt` & `main-mcu-evt`).
+`main-mcu-bootloader` & `main-mcu-app`).
+
+:::caution
+
+It is not recommended to use Software Types to distinguish between software
+build _variants_ (i.e. `main-mcu-development` & `main-mcu-prod`), because this
+does not play nicely with Memfault's concept of "Primary Software Type" (see
+below). We recommend encoding build variants into the Software Version instead,
+by suffixing the version (i.e. `1.0.0-development`).
```
noahp, line 66: makes me think more and more we should require semver 2.0 versions by default, and opt-in to natsort :thinking: 

in semver 2.0, the `-development` field is the "pre-release" indicator. Precedence computation is a little more subtle that major.minor.patch in semver 2.0

PR #566, opened 2023-09-14
feat(mcu-reboots): MFLT-10017 Clarify reboot reason precedence

gminn:
### Summary

 Reboot reasons can be set either before a reboot with
 `memfault_reboot_tracking_mark_reset_imminent` or after a reboot with
 `memfault_reboot_reason_get` and `memfault_reboot_tracking_boot`. Those
 reasons marked before an imminent reboot take precedence over the
 reason found on boot with the platform-specific implementation of
 `memfault_reboot_reason_get`. This fact is now made clear with a note
 calling it out, and a few other edits to differentiate between these two
ways of setting the reboot reason.

 ### Test Plan

Manually verified UI


docs/mcu/reboot-reason-tracking.mdx
```diff
@@ -67,41 +99,22 @@ void ota_finalize_and_reboot(void) {
 }
 ```
 
+:::note Reboot Precedence
+
+On boot, `memfault_reboot_tracking_boot` will always be called, but if a reboot
```
noahp, line 104: OT, what do you think about using `memfault_reboot_tracking_boot()` parens when we reference functions? I'm realizing we're wildly inconsistent across the docs on this (though this doc is pretty internally consistent, which is nice :grinning: )?

noahp (approved):
Looks great, thank you!

docs/mcu/reboot-reason-tracking.mdx
gminn: Heavy +1, glad you feel this way too! I'm on it

gminn: Fixed [here](https://github.com/memfault/memfault-docs/pull/566/commits/1974420d28b3bc75400d53ad8052dbedbdd78dfe)

PR #567, opened 2023-09-14
remove/ Microsoft OAuth

Natasha-Co:
remove "Microsoft OAuth" as an option


ejohnso49:
@chrisc11 Could you double-check the terms used here? We wanted to be sure that we the edits right

chrisc11 (approved):

PR #568, opened 2023-09-18
Document manual upload of Linux coredumps

martijnthe:
- Added new "Uploading Linux Diagnostics" page, which is based on the Android one.
- Couple small tweaks to "Uploading Android Diagnostics" page. Updated the screenshot / .gif.


docs/linux/uploading-linux-diagnostics.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+## Through the Linux SDK / `memfaultd`
+
+When [configured with a Memfault Project Key][integration-guide], `memfaultd`
+will by default, automatically start uploading device diagnostics information
+into Memfault.
+
+During development, it makes sense to configure the `memfaultd` to send data at
+a higher frequency. You can do this by enabling [Developer
+Mode][developer-mode].
```
sarfata, line 17: I don't think that is true. We don't upload at higher frequency in developer mode. You might want to suggest `memfaultctl sync` instead.

docs/linux/uploading-linux-diagnostics.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+[linux-coredumps]: /docs/linux/coredumps
+
+To upload a MAR file that is located at, for example `/path/to/device-data.mar`,
+run the appropriate command:
+
+:::note
+
+The Software Type, Software Version, Hardware Version, and Device Serial
+parameters of the CLI command below should match that of the device which
```
sarfata, line 57: Is that really true? They describe the "uploading device", no? Why do you think this is useful to mention here?

docs/linux/uploading-linux-diagnostics.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+The following examples will use an Organization Auth Token. Please refer to the
+[Memfault CLI Authentication](/docs/ci/install-memfault-cli/#authentication)
+section for more information about the different forms of authentication.
+
+### Uploading a MAR file
+
+A Memfault Archive File (MAR) is an archive format that is generated by
+`memfaultd` containing many files in a compressed ZIP archive. The rationale
+behind creating and using a bundle containing many files is:
```
sarfata, line 39: As one customer recently pointed out to me, we do not publicly document the MAR file format. So I am not sure how useful this will be actually.
But for the sake of keeping our documentation parallel 👍

docs/linux/uploading-linux-diagnostics.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+### Manually capturing a coredump
+
+To manually capture a coredump, you can use one of these techniques:
+
+- You can use [the `gcore` program][gcore-program] (part of the GDB package) to
+  capture a coredump from any running program, given its PID. This program is a
+  script that attaches GDB to the program with given PID and runs [the gcore or
+  generate-core-file GDB command][gcore-gdb-cmd].
+- You can make the kernel capture a coredump of a running program:
```
sarfata, line 101: Maybe worth mentioning also that if the processes crashes, a coredump will be captured automatically (as long as pattern and ulimit allow it).

docs/linux/uploading-linux-diagnostics.mdx
```diff
@@ -0,0 +1,140 @@
...
+
```
sarfata, line 102:
```suggestion
- You can let the kernel capture a coredump of a program:
(...)
\# Wait for program to crash or
\# To force the program to crash, send SIGQUIT signal which...

```

sarfata (approved):

docs/linux/uploading-linux-diagnostics.mdx
martijnthe: Whoops, copy pasta. Thanks.

docs/linux/uploading-linux-diagnostics.mdx
martijnthe: Copy pasted it from the Android page. I think it makes sense to make it match (that's what happens in uploads from the SDK too). I don't think it matters all that much though, because the metadata from the entries should get used for most things.

PR #574, opened 2023-10-04
ESP32 doc refresh

noahp:
Also enable the `cmake` prismjs syntax highlighting, which impacts a few
other spots (we were already using the "cmake" code fence language
identifier in at least one other spot, it just wasn't actually
highlighting, so now it is!)


docs/mcu/espressif-esp32-guide.mdx
```diff
@@ -50,119 +37,117 @@ SDK, under
...
+### 1. Clone Memfault SDK
 
 Using a [Git](https://git-scm.com/) client, clone the `memfault-firmware-sdk`
-repository from:
+repository:
 
-```
-https://github.com/memfault/memfault-firmware-sdk.git
+```bash
+git clone https://github.com/memfault/memfault-firmware-sdk.git
```
noahp, line 63: we might want to suggest adding it as a git submodules to the customer's app project, dunno

docs/mcu/espressif-esp32-guide.mdx
```diff
@@ -50,119 +37,117 @@ SDK, under
...
-### Add Memfault Build ID to the ESP-IDF CMake Build System
+:::
 
-It is recommended that you use a build ID to uniquely identify each build. Do
-this by including Memfault build ID generation by adding a custom post-build
-step to your top level `CMakeLists.txt`.
+The end result should look something like this:
 
-```c
+```bash
```
ejohnso49, line 86: NB: Github's code highlighter package actually supports CMake! Docusaurus' bundled one does not sadly

docs/mcu/espressif-esp32-guide.mdx
```diff
@@ -50,119 +37,117 @@ SDK, under
...
 # CMakeLists.txt (top-level)
-# [...]
+cmake_minimum_required(VERSION 3.16)
 
-# Be sure to set this variable appropriately
+# Item 1 to add for Memfault
```
ejohnso49, line 90: I think a TODO would help highlight this as a required modification. Reader could then search the file for TODO to confirm

docs/mcu/espressif-esp32-guide.mdx
```diff
@@ -243,40 +231,29 @@ partition CSV file". Optionally, you can edit the sdkconfig file directly
...
-#### Metrics and Trace Definitions
-
-The [device metrics](/docs/mcu/metrics-api) and
-[trace event](/docs/mcu/trace-events) subsystems require configuration files to
-be added to your include path (`memfault_metrics_heartbeat_config.def` &
-`memfault_trace_reason_user_config.def`, respectively). Because you copied them
-to your `main` directory these files' locations should be known to CMake.
-
-#### `memfault_platform_get_device_info()`
+### 6. Implement `memfault_platform_get_device_info()`
```
ejohnso49, line 234: We should refresh this after the built-in implementation is released

docs/mcu/espressif-esp32-guide.mdx
```diff
@@ -341,22 +317,22 @@ In ESP-IDF this can be set in a few different places:
...
 
 Assuming that your project has already been built before adding Memfault you
 will need to rebuild now that the Memfault SDK has been integrated into your
 ESP-IDF project.
 
 ```bash
-$ idf.py clean
-$ idf.py menuconfig  # Verify your settings
-$ idf.py build
+idf.py clean
```
ejohnso49, line 330: Should we recommend a `fullclean` instead? I dunno we've both been bitten by the inconsistent clean behavior...

ejohnso49 (approved):
Thanks!

docs/mcu/espressif-esp32-guide.mdx
noahp: ooo we actually have it, but had to add it to the list [here](https://github.com/memfault/memfault-docs/blob/6667a00b420d95f81b72c524144566dd67a26bb4/docusaurus.config.js#L28)! enabled now. great callout.

docs/mcu/espressif-esp32-guide.mdx
noahp: yeah, probably a good idea. i think `fullclean` still doesn't wipe `sdkconfig`, but these instructions are intended to apply to project that just use `sdkconfig.defaults` as well as ones that version their `sdkconfig` file directly, so it should work correctly.

noahp: ah yeah, reading the docs, `fullclean` will wipe cmakelists cached files (it deletes the whole build directory) but leaves `sdkconfig` alone. I'll change it, thank you!

docs/mcu/espressif-esp32-guide.mdx
noahp: the nice thing about a submodule (for advanced users) is that it's tracked with their application git repo, so it's reproducible from a single `git clone --recursive`. it's also what i did in the standalone example for pulling in the Memfault SDK. all moot once we get a esp-idf component shipped though :grin: :tm:

PR #575, opened 2023-10-04
MFLT-11298 Document the changes introduced by 1.6.0 and 1.7.0

sarfata:
### Summary

Document features added in memfault SDK for Linux 1.6 and 1.7:

 - systemd is not required anymore
 - upload_interval_seconds can be set to 0 to disable periodic uploads
 - we use rustls by default and users can revert back to openssl if they want
 - we added a `memfaultctl export` command

PR #576, opened 2023-10-04
doc(linux): first pass at documenting coredump filtering

sarfata:
### Summary

Document an upcoming feature: coredump memory filtering on Linux.


sarfata:
* main
  * **PR #575** <a href="https://app.graphite.dev/github/pr/memfault/memfault-docs/575" target="_blank">Graphite</a> 
    * **PR #576** <a href="https://app.graphite.dev/github/pr/memfault/memfault-docs/576" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/memfault/memfault-docs/576?utm_source=stack-comment).

docs/linux/reference-memfaultd-configuration.mdx
```diff
@@ -236,9 +237,39 @@ periodically (see the top-level `upload_interval_seconds`).
 
 #### Capture Strategy
 
-The only strategy supported at this time is `kernel_selection`. This strategy
-captures all the memory regions included in the ELF core file generated by the
-kernel.
+`memfaultd` can use different strategy to capture the memory of the program
+which crashed.
```
ejohnso49, line 241:
```suggestion
`memfaultd` can use different strategies to capture memory of the crashed program.
which crashed.
```

docs/linux/reference-memfaultd-configuration.mdx
```diff
@@ -236,9 +237,39 @@ periodically (see the top-level `upload_interval_seconds`).
...
+
+By default, the `threads` strategy is used and will only capture the stacks of
+all the threads in the program. This keeps the coredump to a small size and
+provides most of the information that is useful for debugging.
```
ejohnso49, line 245:
```suggestion
By default, the `threads` strategy is used and limits capture to all program thread stacks. This reduces the coredump size but still keeping the most useful information when debugging.
```

docs/linux/reference-memfaultd-configuration.mdx
```diff
@@ -236,9 +237,39 @@ periodically (see the top-level `upload_interval_seconds`).
...
+
+You can limit the size of each stacks with `max_thread_size_kib`.
```
ejohnso49, line 247:
```suggestion
You can limit the size of each stack with `max_thread_size_kib`.
```

ejohnso49 (approved):

docs/linux/reference-memfaultd-configuration.mdx
martijnthe, line 241:
```suggestion
that crashed.
```

Not a native speaker, but I think it should be "that" in this case?
https://www.grammarly.com/blog/which-vs-that/

docs/linux/reference-memfaultd-configuration.mdx
```diff
@@ -236,9 +237,39 @@ periodically (see the top-level `upload_interval_seconds`).
...
+      "type": "threads",
+      "max_thread_size_kib": 32
+    }
+  }
+}
+```
+
+If you would like to capture more memory, and include the heap and other memory
+regions, you can use the `kernel_selection` strategy. This strategy captures all
+the memory regions included in the ELF core file generated by the kernel.
```
martijnthe, line 262: Should we provide a link here to the kernel settings?
We have this in an existing doc page:
https://docs.memfault.com/docs/linux/coredumps/#linux-kernel-configuration

Also wonder whether we should warn here that with this option, the coredumps can grow as large as the amount of memory that the program is using. In case the uncompressed size exceeds `coredump_max_size_kib`, the coredump will not be saved.

martijnthe (approved):

docs/linux/reference-memfaultd-configuration.mdx
sarfata: Awesome reference thanks.

docs/linux/reference-memfaultd-configuration.mdx
sarfata: Yeah agreed. And your comment made me realize that this long explanation would be better in the coredumps guide so I have moved this block there and only kept the syntax here.

bahildebrand (approved):

PR #584, opened 2023-10-12
MFLT-11828 Explicitly mention TMPDIR in  example

patwolfe:
## Summary 
Customers have asked to be able to modify the tmp directory used by 
`memfault upload-yocto-symbols` even though this behavior is already 
supported via `TMPDIR`. Include a tip in the docs to call this out so 
customers are more aware this is an option. 

## Docs 
[MFLT-11828](https://memfault.myjetbrains.com/youtrack/issue/MFLT-11828)
## Test Plan
Screenshot 2023-10-12 at 11 04 24 AM


docs/linux/coredumps.mdx
```diff
@@ -218,6 +218,24 @@ $ memfault \
...
+  environment variable&nbsp;
+  <a href="https://en.wikipedia.org/wiki/TMPDIR">TMPDIR</a>. Just prefix the <code>
+    memfault
+  </code> command like so <code>
+    TMPDIR=&lt;desired directory&gt; memfault ...
+  </code> or&nbsp;
+  <code>export TMPDIR=&lt;tmp driectory&gt;</code> in your shell before running
+  the Memfault CLI command. This works for other CLI commands that use a
+  temporary directory as well!
+</p>
```
sarfata, line 237: I think only a few people will need this so I would keep it a bit shorter.

```suggestion
<p>
  By default <code>upload-yocto-symbols</code> will use a default temporary
  directory in your system such as <code>/tmp</code> to extract the
  debug symbols. This can use significant space and if you'd like to use a
  different location in your filesystem, you can do so with the   <a href="https://en.wikipedia.org/wiki/TMPDIR"><code>TMPDIR</code></a> environment variable&nbsp;:
  <code>
    TMPDIR=&lt;desired directory&gt; memfault ...
  </code>.
</p>
```

sarfata (approved):

PR #592, opened 2023-10-19
Fetch latest NCS version in manifest snippet

gminn:
### Summary

The latest NCS version as well as the latest Memfault SDK version
are not being used in a suggested `west.yml` snippet. Now, the
NCS latest version tag will be pulled, and we can reuse the existing
`ZephyrManifest` function to do this.

Bonus: removed an old, unused zephyr manifest file.

### Test Plan

Verified visually with local server.

Screenshot 2023-10-19 at 7 00 41 PM

Checked that fallbacks work by forcing an error:

Screenshot 2023-10-19 at 6 59 46 PM


noahp (commented):
Brilliant, I love it!

PR #597, opened 2023-10-20
Update Yarn's lockfile

tyhoff:
I'm surprised my Yarn updates this file, but it was required for me to run the test instance locally. Can someone check this?


noahp:
The current yarn.lock file works on my setup :thinking: 
```bash
❯ yarn --version
1.22.19
```

what errors did you get locally?

tyhoff:
Whoa - I was using Yarn 2 or 3 - and couldn't get out of it. Just checked again and now I'm using the same version as you. It all works again. 

It would just rewrite the whole yarn.lock file.

tyhoff:
Not needed

PR #598, opened 2023-10-20
Remove getting-started table from each page

tyhoff:
We have a reminder on pretty much every 'Subsystem Guide' that the customer should complete the 'Getting Started Guide'. It's a large table and it consumes a lot of space.

Instead of the table, make it just a link.


tyhoff:
@noahp It probably doesn't go the full distance, but do you think this is good enough for now?

docs/mcu/arm-infineon-modustoolbox-guide.mdx
```diff
@@ -2,6 +2,7 @@
 id: arm-infineon-modustoolbox-guide
 title: Infineon ModusToolbox™ Getting Started Guide for PSoC™ 6
 sidebar_label: Infineon ModusToolbox™ SDK for PSoC™ 6
+description: "Integrate Memfault using Infineon's ModusToolbox on the PSoC™ 6"
```
tyhoff, line 5: Added this because it was missing from the page showing all the getting started guides.

PR #606, opened 2023-10-24
feat: Organizes Android behind User Guides and Subsystem Guides

Blisse:
Screenshot 2023-10-26 at 11 07 22

Screenshot 2023-10-26 at 11 07 44


sjp4 (approved):
I would probably put SDK reference at the top level, then we need to start breaking up the other big pages.

PR #611, opened 2023-10-25
Update arm-cortex-m-guide.mdx

b1t3s1z3:
Split the document up into 4 sections, re-ordered some items from early in the guide into later sections to reduce the number of things necessary to get some initial value from Memfault:

    - In “Add Test Commands to CLI”, advance “Test different crash types where a coredump should be captured” to beginning of the Stage 2 (before “Define First Trace Event”). Same with Post Data to Cloud[…] and Upload Symbol File.
    - Move Check Exception Handler Implementations to Stage 2 (within Add Test Commands to CLI)
    - Registering the Assert Handler gets deferred to “Fine Tuning” (Step 4)
    - Defer Reboot Tracking, Logging, Timer Dependencies, as well as RTOS Port Files to later section.
    
    
    @noahp I didn't incorporate all the edits that we discussed, and I think the copy that introduces each section could use some work, but can you take a look at this and confirm whether this is now the order that we discussed?


b1t3s1z3:
Also @noahp this is my first PR here at memfault, so let me know if I'm breaking some important conventions here with this (probably low-impact) change that is proposed.

noahp (approved):
Huge fan of this! Just a few minor tweaks to the language needed

docs/mcu/arm-cortex-m-guide.mdx
```diff
@@ -43,7 +43,7 @@ or TI ARM Compiler.
 
 <ProjectAwareCreateAProject />
 
-## Integration Steps {#integration-steps}
```
martijnthe, line 46: The anchor `#integration-steps` is now gone.
There's an [existing mflt.io short link to this anchor](https://github.com/memfault/memfault/blob/c17e8ed0a2186b3a9f5e3bc90264de822b0eaab5/mflt.io/_redirects#L154-L154).

Options I see:
1. Add the `#integration-steps` anchor to the "Stage 1" heading. Mildly confusing because the anchor id doesn't match the title of the heading. 
2. Rename the anchor and the short link (and break any other usages of the old anchor).

I'm leaning slightly towards 2.
Existing uses of the old anchor would still end up on this page, so it's not a totally broken link.

docs/mcu/arm-cortex-m-guide.mdx
```diff
@@ -57,27 +57,56 @@ or TI ARM Compiler.
...
+[I] HW version: dvt1
+[I] Reset Reason, RESETREAS=0x4
+[I] Reset Causes:
+[I]  Software
+[I] Memfault Initialized!
+```
+
+## Stage 2: Get your Data into Memfault
+
+### Reboot Tracking Dependencies
```
martijnthe, line 89: I'm in favor of keeping all the explicit anchor IDs. That way one can freely change the title w/o breaking existing links that include an old, implicit/auto-generated anchor ID.

```suggestion
### Reboot Tracking Dependencies {#reboot-tracking-dependencies}
```

docs/mcu/arm-cortex-m-guide.mdx
b1t3s1z3: Apologies @martijnthe , I don't know enough about how anchors work - what do you recommend we rename it to (does it "matter")? And where does that renaming take place - in the mflt.io/_redirects, and in the arm-cortex-m-guide.mdx?

martijnthe: After your PR, @noahp merged a change that enforces having explicit anchors: https://github.com/memfault/memfault-docs/pull/620

If you merge from master, this PR should also pick up that change.

You can also run `yarn run docusaurus write-heading-ids . docs/*` locally to auto-generate the anchors.

> And where does that renaming take place - in the mflt.io/_redirects, and in the arm-cortex-m-guide.mdx?

Right. In the .mdx, you add the anchor id after the heading title, wrapped in `{#anchor-id-here}`.
And then in the mflt.io/_redirects, you change the link to have a `#anchor-id-here` at the end.

lmk if you need more help with this!

b1t3s1z3: I merged from main to my branch, but now it actually looks like the anchors are being removed by my pr - am I reading that right?

Screenshot 2023-10-31 at 3 39 12 PM

b1t3s1z3: @martijnthe sorry for the confusion, I'm cancelling this PR and creating a fresh one here: https://github.com/memfault/memfault-docs/pull/637

PR #615, opened 2023-10-26
Fix many internal dead links & anchor links

tyhoff:
```
# in one terminal
yarn build
cd build/
python3 -m http.server 3000 

# in another terminal
muffet --exclude 'https://memfault.com/.*' --exclude linkedin.com --exclude github.com --exclude googletagmanager.com --exclude algolia.net --exclude twitter.com --exclude 'https://api-docs.memfault.com/.*' --buffer-size=16384 -c 16 --verbose http://localhost:3000
```

It has to be properly built because it needs to render as static HTML, not dynamically rendered.


sjp4:
This will need to redirects to be updated to the new anchors e.g. https://github.com/memfault/memfault/blob/dab8dc2e6cbbff280dee9c08b99bb5d9724b9deb/mflt.io/_redirects#L34-L33

tyhoff:
Great point - thank you!

sarfata:
wow ... thank you!

PR #618, opened 2023-10-27
SDK version tab for __assert_func hook

noahp:
This is now built-in to the SDK, add a note.


tyhoff:
> Memfault advises against calling assert() directly from a project. Instead, we suggest adding a YOUR_PLATFORM_ASSERT(x) macro as described above instead.

This phrase on 1.4.1+ is no longer valid - there's no `YOUR_PLATFORM_ASSERT` "above" anymore.

tyhoff (approved):
Generally fine with the change!

Just need to fix the note about the `YOUR_PLATFORM_ASSERT` being "above" when it isn't there anymore.

noahp:
> > Memfault advises against calling assert() directly from a project. Instead, we suggest adding a YOUR_PLATFORM_ASSERT(x) macro as described above instead.
> 
> This phrase on 1.4.1+ is no longer valid - there's no `YOUR_PLATFORM_ASSERT` "above" anymore.

ah yep, this needs to be `MEMFAULT_ASSERT()` :call_me_hand: will fix!

ejohnso49 (approved):

tyhoff (approved):

PR #619, opened 2023-10-27
Update with new memfault CLI commands from 1.0.4

patwolfe:
Update docs to account for two new commands, `upload-elf-coredump` and `upload-elf-symbols`. In particular, the current docs imply that non-Yocto builds need to use the API or manually upload ELF symbols which is now no longer the case.


phoenixuprising (approved):
LGTM

PR #621, opened 2023-10-27
Update nRF Connect SDK guide

gminn:
### Summary

Ran through a simple integration using the nRF sample integration
project and made edits along the way.

Found it a bit annoying to find where the sample integration lived
so I just linked to the source up-front.


docs/mcu/nordic-nrf-connect-sdk-guide.mdx
```diff
@@ -486,7 +503,7 @@ uart:~$ mflt get_core
 
 On nRF91 you can post directly to the cloud:
 
-```bash
```
noahp, line 489: hmm, we should be consistent. the `bash` highlighting is not great, probably `plaintext` for the console. i'll put up a change

gminn: Great fix, thank you Noah!

PR #625, opened 2023-10-27
Rework sidebars.js for multiple sidebars

ejohnso49:
Our sidebar is getting quite long because it contains all pages from all categories. This change will split the sidebar into multiple sidebars! This is supported oob by docusaurus and is also how they structure their docs page as well!

The new sidebars are divided between:

- Platform
  - Contains Web App, CI/CD, Best Practiaces, and Reference categories
- MCU
- Android
- Linux


tyhoff:
I notice that Docusaurus does this on their site too. For this to work, it almost needs another header or callout which section is selected. 

Something like 
image
  or 
image

martijnthe:
In the narrow layout (i.e. mobile), it's not very clear that one is "inside" a category, let alone which one:

image

After clicking the hamburger menu:

image


Before, the top level hierarchy was at least shown in the hamburger menu.
I think it's pretty confusing now because you don't have a sense of where you are in the grand scheme of things.

Not sure if this is easy to fix.

ejohnso49:
@martijnthe I think I've got a relatively simple change to help highlight what section is active. There is an [existing problem](https://docs.memfault.com/docs/category/getting-started-guides) where the active navbar item is not styled correctly that I'll look into fixing later. Here's an [example](https://deploy-preview-625--memfault-docs.netlify.app/docs/category/getting-started-guides) with this change. If this is good, then I think we can make this change

ejohnso49:
Some screenshots with details.

Full screen Sidebar/Navbar:
image

Mobile Sidebar/Navbar:
image

src/css/custom.css
```diff
@@ -142,18 +142,28 @@ section.features_src-pages- {
...
+
+/**
+ * Add a bottom border to emphasize which sidebar section is active
+ * This helps highlight what section of the docs the pages in the sidebar belong to
+ */
+.navbar__link--active,
+.navbar__link--active:hover,
+.theme-doc-sidebar-item-category-level-1 > div > a.menu__link {
+    text-decoration: underline 4px;
+}
```
tyhoff, line 169: 🪄  Amazing. Love it.

tyhoff (approved):
🚢  This is awesome!

sidebars.js
```diff
@@ -203,50 +269,5 @@ module.exports = {
                 "linux/cross-compile-memfaultd",
```
patwolfe, line 269: Nit: `cross-compile-memfaultd` just got moved up to be immediately after the Yocto integration guide in #645. (This looks awesome otherwise!)

src/css/custom.css
```diff
@@ -142,18 +142,28 @@ section.features_src-pages- {
...
     border-left: 1px solid var(--ifm-toc-border-color);
 }
+
+/**
+ * Add a bottom border to emphasize which sidebar section is active
+ * This helps highlight what section of the docs the pages in the sidebar belong to
+ */
+.navbar__link--active,
+.navbar__link--active:hover,
+.theme-doc-sidebar-item-category-level-1 > div > a.menu__link {
```
Blisse, line 167: super nit and i'll probably play around with it after, but on the sidebar it'd look better if the underline spanned the whole section and not just the text element~

docusaurus.config.js
```diff
@@ -45,7 +45,7 @@ module.exports = {
                 {
                     to: "docs/platform/introduction",
                     activeBasePath: "docs/platform",
-                    label: "Web App",
+                    label: "Platform",
```
Blisse, line 48: This is the best part of the change TBH :D

Blisse (approved):

src/css/custom.css
ejohnso49: Yeah, would welcome any help with the styling ❤️ . This was definitely a minimal effort/change here

sidebars.js
ejohnso49: Good callout, I'll make sure I rebase before I wrap this up

docusaurus.config.js
noahp: @fnune @HBehrens FYI, we went back and forth on this before so want to make sure you see it before we update

martijnthe (approved):

src/css/custom.css
fnune: image

Tips/complaints:

1. Why is the underline necessary also here? I'd leave it out
2. Actually, I'd remove this whole menu item (in the screenshot `MCU`) since it only leads to a cards-view of the top-level items, which are also shown in the sidebar
3. I'd use less pickles for that underline :)

fnune: My suggestions:

image

- Remove the `text: underline` changes, there's already enough context to tell which top-level menu item is active
- Remove the top-level `generated-index` items and leave each sub-sidebar a flatter list of things

Here's a PR based on your PR: https://github.com/memfault/memfault-docs/pull/657

fnune: If you really want an underline, here's a hacky suggestion:

```css
.navbar__link {
    position: relative;
}
/**
 * Add a bottom border to emphasize which sidebar section is active
 * This helps highlight what section of the docs the pages in the sidebar belong to
 */
.navbar__link--active::after {
    --highlight-size: 3px;

    content: "";
    position: absolute;
    top: calc(50% + (var(--ifm-navbar-height) / 2) - var(--highlight-size));
    width: 100%;
    left: 0;
    height: var(--highlight-size);
    background-color: var(--ifm-navbar-link-hover-color);
    border-radius: 2px 2px 0 0;
}
```

image

Although `text-decoration: underline 2px` also looks OK IMO.

fnune: Ah shoo, I realized I'm suggesting undoing your fix for mobile. I'm not sure, I feel like the `generated-index` entities look really out of place on desktop.

Sorry to add to the confusion, I'll `:disappear:`. Take whatever you prefer :)

martijnthe:
@fnune 

> Actually, I'd remove this whole menu item (in the screenshot MCU) since it only leads to a cards-view of the top-level items, which are also shown in the sidebar

See [my comment here](https://github.com/memfault/memfault-docs/pull/625#issuecomment-1784900039). I think leaving it out is pretty confusing on mobile, where you don't see the top level nav.

fnune:
> @fnune
> 
> > Actually, I'd remove this whole menu item (in the screenshot MCU) since it only leads to a cards-view of the top-level items, which are also shown in the sidebar
> 
> See [my comment here](https://github.com/memfault/memfault-docs/pull/625#issuecomment-1784900039). I think leaving it out is pretty confusing on mobile, where you don't see the top level nav.

Yup, I realized that later in the same thread: https://github.com/memfault/memfault-docs/pull/625#discussion_r1384716025

PR #635, opened 2023-10-31
TLS 1.2 cipher suite support

noahp:
https://support.memfault.com/hc/en-us/articles/14497708585885-Which-TLS-1-2-cipher-suites-are-supported-on-the-Memfault-device-endpoints-


docs/reference/reference-fqdn.mdx
```diff
@@ -12,7 +12,7 @@ is sent over SSL.
 
 :::
 
-### Overview
+## Overview {#overview}
```
tyhoff, line 15: Gahh I broke my own rule!

tyhoff (approved):

PR #648, opened 2023-11-03
Global reflink plugin

phoenixuprising:
Create a custom remark plugin that reads in a file with a list of reflinks which can be refrerenced from any markdown file in the instance.


src/remark/global-ref-links/index.js
```diff
@@ -0,0 +1,59 @@
+const visit = require("unist-util-visit");
+const util = require("util")
+
+const plugin = (options) =>  {
```
phoenixuprising, line 4: Don't judge me for any of this, literally the first time I've written any `js` in my life. I will happily take any feedback on this, I will not be offended.

src/remark/global-ref-links/index.js
```diff
@@ -0,0 +1,59 @@
...
+        
+        if (errors.length !== 0) {
+            errors.forEach((error) => {
+                console.error(error)
+            })
+            throw new Error(util.format("refs declared multiple times. Remove duplicate refs from %s", linksFile))
+        }
+
+        // Actually transform the AST node from a `linkReference` to a `link` with the correct url.
+        visit(ast, "linkReference", (node) => {
```
phoenixuprising, line 46: Ideally, this would be the only code within the `transformer` but I couldn't figure out how to pass in the `refs` dictionary into the it.

src/remark/global-ref-links/index.js
```diff
@@ -0,0 +1,69 @@
...
+                errors.push(util.format("%s - linkref definition %s found in global-ref-link file %s", mdFile, node.label, linksFile))
+            }
+        })
+        hasErrors(errors, util.format("local linkref definitions in %s matching those found in global linkref file %s", mdFile, linksFile))
+
+        // Actually transform the AST node from a `linkReference` to a `link` with the correct url.
+        visit(ast, "linkReference", (node) => {
+            if ( refs[node.label] !== undefined ) {
+                let url = refs[node.label]
+                console.log(util.format("%s: replacing %s with %s", mdFile, node.label, url))
```
phoenixuprising, line 59: should probably comment this out, it will get *very* noisy long term.

PR #651, opened 2023-11-05
Check `curl` exit code in generic OTA script for Linux

patwolfe:
Adds a check on the `curl` exit code in the OTA agent script for Linux that will differentiate errors caused by failing to install or download the OTA payload from errors communicating with the Memfault OTA API in the first place. 

Also updates the styling on the bash code block.

Updated look:

Screenshot 2023-11-05 at 6 22 22 PM


ejohnso49 (approved):

sarfata (approved):
🙏

PR #655, opened 2023-11-06
Update Android sync section with table and subsections

ejohnso49:
### Summary

Made some changes based on support tickets we received. Main goal was to emphasize what the different parameters are used for


docs/android/android-bort.mdx
```diff
@@ -528,20 +528,50 @@ collected from the system, it is uploaded up to 15 minutes later.
...
 (if true) or
 [NetworkType.UNMETERED](https://developer.android.com/reference/androidx/work/NetworkType#UNMETERED)
 constraint on its network requests.
 
-Metrics, logs, and DropBoxManager Entries are uploaded every 2 hours by default.
-OTAs are checked for every 12 hours. Both settings can be adjusted if desired by
-contacting support.
+#### Sync Rates {#sync-rates}
+
+The table below outlines the default rates for how often each type of Memfault
```
sjp4, line 548: I would actually probably revert this change - I'm not sure it makes sense to break out metrics/logs/dropbox when they are all uploaded as one `mar` bundle every 2 hours.

sjp4 (approved):

docs/android/android-bort.mdx
ejohnso49: That's a good idea, I can do that! I can put up a change to summarize that

PR #657, opened 2023-11-07
Fausto's preference

fnune:


noahp:
sweet, i like this!

tyhoff:
Do we need to keep this open or can we close it now? (I don't really know what it is)

fnune:
This is essentially an ellaborate code suggestion I had for @ejohnso49 on top of https://github.com/memfault/memfault-docs/pull/625

We can close it, it's always here if we want to take it up.

PR #661, opened 2023-11-08
Fix domain name in generic OTA script for Linux

mnaberez:


tyhoff (approved):
Well hello there! Appreciate the fix!

PR #666, opened 2023-11-10
Remove mentions of firmware SDK 0.3x or below

tyhoff:
These versions are over a year old, and no one should be integrating with them anymore.

Also update the Zephyr supported versions from 3.3 to 3.5 since I think we support it!


noahp:
Nice update! We are compatible with zephyr 3.5 👍

noahp (approved):

PR #680, opened 2023-11-20
Add guide: "Debugging Linux coredumps with GDB"

martijnthe:
Started a new guide "Debugging Linux coredumps with GDB".
It's based on the MCU equivalent, but trimmed down a bit.

I removed the sections about our conda gdb distros, because I don't have 100% confidence in that they will work correctly.


docs/linux/linux-coredump-elf-with-gdb.mdx
```diff
@@ -0,0 +1,101 @@
...
+  width="285"
+  src="/img/docs/linux/download-coredump-elf.png"
+  alt="/img/docs/linux/download-coredump-elf.png"
+></ImageFigure>
+
+## Configure `DEBUGINFOD_URLS` {#configure-debuginfod-urls}
+
+Aside of the ELF core file, GDB also needs debug information about the software
+that generated the coredump, in order to make sense of the data in the core
+file. This debug information is present in the symbol ELF files.
```
sarfata, line 40:
```suggestion
In order to make sense of the data in the ELF core file, GDB also needs debug information about the software
that generated the coredump. This debug information is present in the symbol ELF files.
```

docs/linux/linux-coredump-elf-with-gdb.mdx
```diff
@@ -0,0 +1,101 @@
...
+If your version of GDB does not have support for core files, you will see this
+message when attempting to load the file:
+
+```
+"/path/to/coredump.elf": no core file handler recognizes format
+```
+
+Memfault provides GDB binaries that do have support for core files. See
+[section at the bottom](#installing-multi-architecture-gdb-with-conda) for
+installation instructions.
```
sarfata, line 88: This link does not seem to work. I am not sure what was the idea here.

https://deploy-preview-680--memfault-docs.netlify.app/docs/linux/linux-coredump-elf-with-gdb#installing-multi-architecture-gdb-with-conda

sarfata (approved):
Looks great. Thank you.

I think it would be good to mention here what are the limitations imposed by our memory capture strategy because that is not obvious to me. Obviously, not all of the memory will be available. What does this look like in GDB?

docs/linux/linux-coredump-elf-with-gdb.mdx
martijnthe: Ah whoops. Thanks, forgot to remove this.

The gdb v12 still seems to have some bugs around debuginfod which appear to have been fixed in v13.
We haven't updated our conda packages to v13 yet, so I didn't want to point people to it because they're going to have a bad time.

When updating to v13 I'll revisit this doc to add the section back.

PR #684, opened 2023-11-21
Generate CLI docs from the CLI help output

ejohnso49:
## Summary

Our CLI documentation is out of sync with the current release of the CLI. To help with this, I've created a small Python generator script that uses Jinja2 templates to output MDX partials or docs pages composed of the help output from the CLI. This at least gives us something tied to the actual release and an easier process for regenerating these pages in the future. The sidebar sections have been further reworked to better match these changes

### Notes

- I used Python instead of JS because our CLI is already Python based and Click has a very convenient function to get info on the CLI commands
- There are templates for each generator type. The templates can be found in `cli-docs/templates`


cli-docs/cli_doc_gen.py
```diff
@@ -0,0 +1,164 @@
...
+
+def validate_docs_root(docs_root: Path) -> None:
+    ci_docs_path = docs_root / CI_DOCS_PATH
+    template_path = docs_root / TEMPLATE_PATH
+    partials_path = docs_root / PARTIALS_PATH
+    subcommands_docs_path = CI_DOCS_PATH / "cli"
+
+    for path_var in [ci_docs_path, template_path, partials_path, subcommands_docs_path]:
+        if not path_var.exists() or not path_var.is_dir():
+            raise RuntimeError(f"{path_var} does not exist, using {docs_root} as root dir")
```
ejohnso49, line 134: Might be better to create these paths if they don't exist rather than raising an exception?

tyhoff:
This is great.

Make sure to rebase, push those changes, and have a doc somewhere internally about how to rebuild the CLI docs so that we can do this more easily in the future!

I think you should feel free to merge if you feel good about it.

PR #691, opened 2023-11-22
Add DoNotIndex component and add to pages missing from sidebar

ejohnso49:
# Summary

Improve the searchability of our docs by removing these hard to find pages from results


tyhoff (approved):
Nice - this even removes the pages from the `sitemap.xml`.  Love it.

PR #692, opened 2023-11-22
Add note for SWUpdate + Delta Updates

ejohnso49:
Add note to detail how SWUpdate handles delta updates automatically


sarfata:
@ejohnso49 the uploaded payload still needs to be prepared in a specific way (we have internal docs on this but not public yet). I wonder if this statement might lead people to believe that it will work completely automatically. It does not.

vlindhol (approved):

ejohnso49:
> @ejohnso49 the uploaded payload still needs to be prepared in a specific way (we have internal docs on this but not public yet). I wonder if this statement might lead people to believe that it will work completely automatically. It does not.

Thanks for reviewing, I'll close this for now 👍

PR #694, opened 2023-11-27
fix: update broken link

nihalgonsalves:
There was a broken link in #686 (the newline between the text and link caused the text to be interpreted as a standalone anchor). This fixes it.

An upstack PR fixes the CI too.

PR #695, opened 2023-11-27
ci: add yarn build check

nihalgonsalves:
Add a CI build step to prevent issues such as the downstack PR.


nihalgonsalves:
* main
  * **PR #694** <a href="https://app.graphite.dev/github/pr/memfault/memfault-docs/694?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #695** <a href="https://app.graphite.dev/github/pr/memfault/memfault-docs/695?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/memfault/memfault-docs/695?utm_source=stack-comment).

HBehrens (approved):
Thanks for adding the build step to CI.

Minor, but are we now build twice (once here and once for the Netlify deploy)? Could we leverage the output of this new step for the latter to speed up the Netlify deploy without having a build-blocking dependency on Netlify?

nihalgonsalves:
@HBehrens I wouldn't bother; it's fast enough that it isn't a problem and that would make it much more complex – what I will look at is why the preview step passed in the broken PR, it should also fail. 

It may have been due to basing off a branch before the plugin PR, and merging after. The plugin PR made an unlinked `[xxx]` block a build-breaking change, before it it was simply an unlinked piece of text.

PR #697, opened 2023-11-27
chore: use lefthook

nihalgonsalves:
We had a pre commit file but in practice that was not used since it required an extra install step.

This now uses lefthook, which is automatically installed by `yarn install`.

Note that we lose the following hooks (the rest are replaced by prettier):

```yaml
- id: check-added-large-files # Prevent giant files from being committed
- id: check-case-conflict # Check for files that would conflict in case-insensitive filesystems
- id: check-executables-have-shebangs # Ensures that (non-binary) executables have a shebang.
```

but I believe this is still better because those were not being enforced on most people's local repos anyway.


nihalgonsalves:
* main
  * **PR #697** <a href="https://app.graphite.dev/github/pr/memfault/memfault-docs/697?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/memfault/memfault-docs/697?utm_source=stack-comment).

utkuoymak (approved):
This is great, thanks! I had to do multiple commits in this repo before just because pre-commit hook wasnt enforced

mrcljx (approved):

mrcljx:
An re: `check-added-large-files` we have `check-large-files` in `lefthook`.

mrcljx (approved):

PR #718, opened 2023-12-05
ci: add nightly automatic PR

nihalgonsalves:
Creates a PR automatically when the CLI help text doesn't match (instead of sending a notification)


.github/workflows/nightly.yml
```diff
@@ -7,44 +7,37 @@ on:
...
+          python-version: "3.10"
+          cache: pip
+
+      - run: yarn
+      - run: pip install --upgrade -r requirements.txt
+      - run: memfault --version
+
+      - run: yarn gen
+
+      - uses: peter-evans/create-pull-request@v5
```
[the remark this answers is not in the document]
nihalgonsalves: nope, see add-paths below. it just ran and deleted #717 (which was created before the add path filter) because there were no changes.

requirements.txt
```diff
@@ -1 +1,2 @@
 memfault-cli
+jinja2
```
noahp, line 2: nit should probably pin this to whatever version we're installing now

.github/workflows/nightly.yml
```diff
@@ -7,44 +7,37 @@ on:
...
-  run-if-failed:
-    runs-on: ubuntu-latest
-    needs: [auto_checks]
-
-    # run if any job failed, only on the scheduled runs
-    if: |
-      always() && contains(needs.*.result, 'failure') && github.event_name == 'schedule'
-    steps:
-      - name: Slack Notification
-        uses: rtCamp/action-slack-notify@v2
```
noahp, line 43: I'd vote keep the slack notification if a PR is created, with `SLACK_COLOR: success`, so someone knows to go and review the change.

nihalgonsalves: It'll auto-request owners-docs, see #717

PR #731, opened 2023-12-11
chore(deps): upgrade docusaurus@v3 node@v20 yarn@v4

nihalgonsalves:


docusaurus.config.ts
noahp, line 1: Can you include in your commit message an explanation of this change (just for our future reference)

phoenixuprising (approved):

PR #748, opened 2023-12-14
FAQ note for IAR malformed debug info workaround

noahp:


ejohnso49 (approved):

docs/mcu/faq.mdx
```diff
@@ -217,3 +217,22 @@ devices that target different projects.
 // chunks or perform OTA requests
 g_mflt_http_client_config.api_key = "YOUR_API_KEY";
 ```
+
+## IAR Symbol Files {#iar-symbol-files}
```
tyhoff, line 221: Is there a specific Chronicler error that would be spit out the user can search for to know they need to do this?

For our future selves, I'm sure we can detect this on symbol file upload - would it be worth pointing to a doc page when a user uploads a bad IAR ELF file?

ejohnso49: I think we can, I can file some details on it

gminn (approved):

PR #757, opened 2023-12-20
docs(metrics): add reboots & crashes docs

nihalgonsalves:
Resolves: MFLT-13050


nihalgonsalves:
* `main`
  * **PR #757** <a href="https://app.graphite.dev/github/pr/memfault/memfault-docs/757?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/memfault/memfault-docs/757?utm_source=stack-comment).

docs/platform/charts.mdx
```diff
@@ -179,6 +179,45 @@ Devices reporting the given value. When visualizing absolute values for
...
+
+### Crashes per 10,000 hours {#crashes-per-10k-hours}
+
+:::info
+
+This chart is currently only available for MCU devices.
+
+:::
+
+The Crashes per 10,000 hours chart is another way to compare the stability of
```
can-memfault, line 208: We don't explain much here and this chart sparked a lot of questions when we shared the first prototype with the customers. We need to explain: 
- Which metric is used to calculate the crash numbers (see the question from [Airthings](https://memfault.slack.com/archives/C01BMCKB17A/p1700475610138909?thread_ts=1700475338.037049&cid=C01BMCKB17A)
(Maybe also the extrapolation, [see Whoop's feedback](https://www.notion.so/memfault/Crashiness-Chart-Feedback-34f00ac34861440ebd6ea1d095502dc4?pvs=4#7e8c10489a5747eaa888a711537b0d6e)
- How do we calculate the operational hours, especially for session based devices: See GoPro's comment in the screenshots below (it's from a private channel) : 
image
image
image


See what @mjessome wrote up before sharing the prototype chart with customers: https://www.notion.so/memfault/Crashiness-Beta-Unexpected-Reboots-7a5d74c0202d4b038f2918db819dbc60

can-memfault: We should also add the public doc link to the info box on the crashiness chart once this goes live

can-memfault: Speaking of that, I think we did a better job explaining this in the info box of the chart: 
image

can-memfault: Also please add since which SDK version these metrics are collected

can-memfault (changes requested):

mjessome (approved):
Looks good to me, though the Crashes / 10K Hours screenshot is a bit misleading because # of hours is below the stated threshold. Would it be possible to update that so operational hours > 5k?

PR #760, opened 2023-12-21
Fix nightly, re-enable slack notif. on failure

noahp:
IMO worth the spam in the channel so we fix it. I get the email
notification but don't want to be the weak link ^_^


noahp:
Manually triggered:
https://github.com/memfault/memfault-docs/actions/runs/7289065023/job/19862921326

